PR #165, opened 2015-02-14
Font color adjustment

### LINKED ISSUE #164: Default font color in dark theme
The base font color lightness is just *ever-so-slightly* too dark and **doesn't provide enough contrast** to the background, IMO.  It makes reading the text somewhat of a chore and turns me off of the otherwise great-looking new dark theme and to me, **that makes it a bug.**

And since I thought it was a problem, **I took a shot at finding a solution.**, I did a little testing and came up with this ***tiny*** adjustment below (pull request incoming).  Original HEX values temporarily **converted to HSL to isolate the issue source** using devtools and show differences easily.  

**Example:**
```css
pre, code, samp, ._maxcdn .url, ._redis>.example, ._haskell .src { ... }
```
Old
-
> color: hsl(48, 3%, 70%);
    = #B4B3AF

Goldilocks-approved
-
> color: hsl(48, 3%, **90%**);
    = **#E6E6E5**

The other colors (syntax highlighting) also need adjusting in the same manner.  I'll try to pull req those too.

*But don't take my word for it.*  
**Try it yourself** in devtools and see if you agree.  *Maybe it's just me.*

My setup:  
- Chrome 40 x64
- ASUS LED monitor
    - tested with all default monitor modes, **sRGB** being lower end of the scale, **Scenery** at the top.

PS - love the new theme though.  It almost matches just about **everything else I use** :+1: :)

dhenson02:
Increased contrast of font against new dark background.  Reference issue #164 for details.

(closes #164)

### DIFF
--- a/assets/stylesheets/components/_prism.scss
+++ b/assets/stylesheets/components/_prism.scss
@@ -18,7 +18,7 @@
 .token.symbol,
 .token.deleted {
   @if $style == 'dark' {
-    color: #cf6a4c;
+    color: #eb8160;
   } @else {
     color: #905;
   }
@@ -31,7 +31,7 @@
 .token.builtin,
 .token.inserted {
   @if $style == 'dark' {
-    color: #d0c584;
+    color: #ddcf88;
   } @else {
     color: #5e8e01;
   }
@@ -43,7 +43,7 @@
 .language-css .token.string,
 .style .token.string {
   @if $style == 'dark' {
-    color: #8f9d6a;
+    color: #b1c676;
   } @else {
     color: #a67f59;
   }
@@ -53,7 +53,7 @@
 .token.attr-value,
 .token.keyword {
   @if $style == 'dark' {
-    color: #7587a6;
+    color: #91b3ed;
   } @else {
     color: #0070a3;
   }
@@ -61,7 +61,7 @@
 
 .token.function {
   @if $style == 'dark' {
-    color: #ac885b;
+    color: #ecbd83;
   } @else {
     color: #dd4a68;
   }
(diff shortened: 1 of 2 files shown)


assets/stylesheets/global/variables-dark.scss
```diff
@@ -9,7 +9,7 @@ $sidebarMediumWidth: 16rem;
 
 $contentBackground: #2d2d2d;
 
-$textColor: #b4b3af;
+$textColor: #e6e6e5;
```
Thibaut: I find this too bright and too close to links (which by the way could you change them to `#fff` [here](

Thibaut: I find this too bright and too close to links (which by the way could you change them to `#fff` [here](https://github.com/dhenson02/devdocs/blob/master/assets/stylesheets/global/variables-dark.scss#L32)).

Does `#c9c9c9` look bright enough on your screen?

assets/stylesheets/components/_prism.scss
```diff
@@ -61,7 +61,7 @@
 
 .token.function {
   @if $style == 'dark' {
-    color: #ac885b;
+    color: #ecbd83;
```
Thibaut: I also find this one too bright. Does `#c79e6b` work for you?

dhenson02: That's a good compromise yeah

assets/stylesheets/global/variables-dark.scss
dhenson02: #c9c9c9 isn't a bad middleground and changing the links to #FFF is more than doable I'll submit that now.

dhenson02:
@Thibaut requested changes applied, see what you think

Thibaut:
Thanks! :+1:

PR #168, opened 2015-02-18
Add docs for Minitest

hollingberry:


Thibaut:
I would prefer adding Minitest as a separate doc:
http://docs.seattlerb.org/minitest/
We just need an icon for it. Any idea?

hollingberry:
I don't think Minitest has an icon. The Seattle Ruby Brigade has an icon at http://www.seattlerb.org:

seattle_rb_logo_big-eea0587cb287f72012cfa249caf120dc

As well as an icon at http://github.com/seattlerb:



Either of these would be fine.

And I agree, adding it as its own doc seems like a good idea because the version of Minitest that comes bundled with Ruby (and that is thus included in its documentation) is usually pretty outdated. Although I think it *was* just updated for the release of Ruby 2.2.

Thibaut:
Minitest was actually removed stdlib in Ruby 2.2, so this PR isn't going to work. I'll try to add Minitest as a separate doc this week-end or the next.

PR #170, opened 2015-02-27
Adding Official Url for url scrapped documentations

### LINKED ISSUE #19: Add links to the official website on each documentation's root page
Related issue: #18

TrangPham:
Closes #19 

This is my first attempt at this. To be honest I only have a slight understanding of what it going on here. Also the styles may need to be changed. @Thibaut Feel free to pick this apart.


Thibaut:
Thanks for your contribution. I think we should use a new attribute (`home_url`) instead of `base_url`, which isn't always a valid URL, and the box should go at the top instead of the bottom.

TrangPham:
@Thibaut Thanks :)

I think I've already made the box go at the top. 
```Ruby
class OfficialUrlFilter < Filter
    def call
      official_url_html << html if base_url
    end
```
https://github.com/TrangPham/devdocs/blob/master/lib/docs/filters/core/official_url.rb#L4
I'm appending the rest of the html to the html of the `official_url_html`. 

Should I rename `official_url` file to `home_url` too?

Thibaut:
> I think I've already made the box go at the top. 

You're right. I read the code wrong.

> Should I rename `official_url` file to `home_url` too?

:+1:

lib/docs/filters/core/official_url.rb
```diff
@@ -0,0 +1,15 @@
+module Docs
+  class OfficialUrlFilter < Filter
+    def call
+      official_url_html << html if base_url
```
Thibaut: This will return `nil` and :boom: when `!base_url`.

TrangPham: Oh you're right - should be
```Ruby
if home_url
  home_url_html << html
else
  html
end
```

Right?

Thibaut: Yep. Or:
```ruby
html.prepend(home_url_html) if home_url
html

**Comment by TrangPham** [comment]
After renaming to home_url - should I be manually adding the urls into each doc's scrapper? 
Also is there a format that the urls should be following?
Should the css class names be changed from `_official` and `_official-link` to home?

**Comment by Thibaut** [comment]
> After renaming to home_url - should I be manually adding the urls into each doc's scrapper? 

I'll add those progressively when I update each doc.

>  Also is there a format that the urls should be following?

Not sure what you mean. The link should go to the project's homepage.

> Should the css class names be changed from `_official` and `_official-link` to home? 

:+1:

File: lib/docs/filters/core/home_url.rb [hunk 2]
```diff
@@ -0,0 +1,16 @@
+module Docs
+  class HomeUrlFilter < Filter
+    def call
+      html.prepend(home_url_html) if home_url
```
Thibaut: You'll need to add a class attribute in `doc.rb`, [`here`](https://github.com/Thibaut/devdocs/blob/dfdd4f373ef42a927c97b0da4d5612893d7b0232/lib/docs/core/scraper.rb#L92-L93), and in `filter.rb` to make this work. Maybe add some tests too.

TrangPham: @Thibaut 

I'm not sure what you wanted changed in `doc.rb`. Is that `lib/docs/core/doc.rb` that you meant?

TrangPham:
I think `_home-link` is already being used as a css class. Should I leave the `_official` and `_official-link` css class named as there are?

TrangPham:
I'm looking through the test failures right now, and will be adding some test cases for home_url.

Do you want home_url to ever be able to be blank? Since you'll be adding them in gradually? Because as I have it now, it will cause `URI::InvalidURIError: bad URI(is not URI?): ` when home_url is blank.

Thibaut:
> Should I leave the `_official` and `_official-link` css class named as there are?

:+1: 

> Do you want home_url to ever be able to be blank? Since you'll be adding them in gradually?

Yes. It should be blank for all docs in this PR.

lib/docs/filters/core/home_url.rb
Thibaut: > I'm not sure what you wanted changed in `doc.rb`. Is that `lib/docs/core/doc.rb` that you meant?

Yes, I meant here:
https://github.com/Thibaut/devdocs/blob/master/lib/docs/core/doc.rb#L7
This is an attribute that all docs will have, scraper or not.

TrangPham:
@Thibaut I think this is ready for review - are there other places which I should add more tests to?

lib/docs/core/scraper.rb
```diff
@@ -34,7 +34,7 @@ def filters
     self.text_filters = FilterStack.new
 
     html_filters.push 'container', 'clean_html', 'normalize_urls', 'internal_urls', 'normalize_paths'
-    text_filters.push 'inner_html', 'clean_text', 'attribution'
+    text_filters.push 'home_url', 'inner_html', 'clean_text', 'attribution'
```
Thibaut: `home_url` needs to happen after `inner_html`, which converts the document to a string (puts it before or after `attribution`).

lib/docs/core/scraper.rb
```diff
@@ -65,6 +65,10 @@ def root_url
       @root_url ||= root_path? ? URL.parse(File.join(base_url.to_s, root_path)) : base_url.normalize
     end
 
+    def home_url
+      @home_url ||= URL.parse self.class.home_url if self.class.home_url
```
Thibaut: No need to parse the URI, we only ever use it as a string. We can just return `self.class.home_url`.

Thibaut:
I'll merge this this week-end.

TrangPham:
@Thibaut Changes made.

And thanks so much for helping me with this PR! I thought it would be easier than it was. I still don't completely understand the mechanics of how devdocs works.

Thibaut:
Didn't do much work on DevDocs this week-end. This will have to wait another week :P

PR #185, opened 2015-03-25
Update ruby.rb to get 2.2.1 documentation.

ab5124:
Changed the version to to the latest release of ruby.

### DIFF
--- a/lib/docs/scrapers/rdoc/ruby.rb
+++ b/lib/docs/scrapers/rdoc/ruby.rb
@@ -10,7 +10,7 @@ module Docs
     #   --no-line-numbers \
     #   --op html .
 
-    self.version = '2.2.0'
+    self.version = '2.2.1'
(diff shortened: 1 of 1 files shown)


Thibaut:
Thanks but unless you've manually checked that the scraper still works, PRs like this aren't really helpful.
https://github.com/Thibaut/devdocs/blob/master/CONTRIBUTING.md#updating-existing-documentations
I'll update the docs over the week-end.

ab5124:
I verified by downloading the docs with Thor. Thanks for updating it though.

Thibaut:
@ab5124 if you ran `thor docs:download ruby` that only fetches the 2.2.0 version I personally scraped and uploaded to DevDocs's server. To generate the docs you need to run `thor docs:generate`.

ab5124:
Alright thanks for that information.  It fails with both 2.2.0 and 2.2.1 when trying to pull the docs from localhost.

PR #205, opened 2015-05-04
Add docs for Drupal

sarhugo:
Drupal docs from http://api.drupal.org


Thibaut:
Thanks! Will take a look this week-end.

Thibaut:
Sorry I didn't look at this yet. Will do it next week-end.

Thibaut:
Thanks again for your contribution & hard work. This is adding a lot of complexity for one set of docs, which makes it harder / more time-consuming for me to maintain things in the future. I'd like to explore ways we can get simplify this before moving forward (unfortunately Drupal has the worst URLs of any docs I've ever seen, by a long margin :/).

Questions & comments:

1. Why did you need the `p.gsub! /\./, '_'` and `p.gsub /__/, '..'` in `fixUri`? (can you provide examples/stacktraces where it failed?)

2. If we add an option to run each URL/path through `URI.decode_www_form` for a given scraper, we should be able to get rid of `fixUri` and the monkey patches in `NormalizePathsFilter`, `InternalUrlsFilter` and `EntriesFilter`. I'll look into that when I have time.

3. Each entry should have a type (never `nil`, unless the doc has no types at all), otherwise they don't appear in the sidebar. Guides &co can be grouped together or skipped (API stuff is more important).

4. Skip all the changelog stuff ("Updates from 6.x to 7.x"). The doc is clocking at 50MB, we should try to trim down the fat as much as possible (more entries == less useful search, bigger size == more likely offline storage will exceed the browser's quota).

5. I'd prefer we remove all the `/7`, e.g., instead of scrapping https://api.drupal.org/api/drupal/includes!form.inc/function/batch_get/7 we should scrape https://api.drupal.org/api/drupal/includes!form.inc/function/batch_get. We could do that using `fix_urls` similarly to [here](https://github.com/Thibaut/devdocs/blob/744d7caa8a504998a14868bf1bb6db416e632ce3/lib/docs/scrapers/laravel.rb#L35).

6. Please remove `public/docs/docs.json` from the commit.

sarhugo:
Hi @Thibaut, thanks for reviewing, I'll also look at this on weekend.. hard week of work :(

sarhugo:
1. Removed the dot replacements at `fixUri`. It works fine without it.
2. Not sure if it will solve the problem. The replacements done was to fix the urls like "https://api.drupal.org/api/drupal/includes!ajax.inc/group/ajax/7" which include an exclamation mark (we have urls with ! + ; : ). If I don't replace these chars I get 404 errors when accessing via the app.
3. I removed a lot of these skipping updates, test, etc. For the remaining, even if they won't appear in the sidebar still are searchable. Lot of these functions are hook implementations of the core modules.
4. Done. Added several skips
5. Done. Whenever drupal v8 is released (and default in docs) we will need to get it back.
6. Done

Thibaut:
Thanks! I have a busy next few weeks coming up; I'll take a look as soon as I can.

Thibaut:
Merged :)
https://github.com/Thibaut/devdocs/commit/dac9b9b03cc31f82e66325fe442fce5b35344d7a
https://github.com/Thibaut/devdocs/commit/79822f8ebc3e5a7b2e1ecb416dd1de191c3c3bc9
http://devdocs.io/drupal/

PR #244, opened 2015-08-12
Update chai version

JaKXz:


Thibaut:
As far as I can tell Chai's website hasn't been updated since 2.1.0. I'll update the DevDocs doc as soon as they do,

PR #286, opened 2015-11-03
Add Bluebird docs

awalGarg:
Heyo! This includes all of the API reference documented at http://bluebirdjs.com/docs/api-reference.html. I left out all the guides in there which are incomplete for now. They can easily be enabled later when bluebird catches up, by removing their entries from the `:skip` array.

/cc @petkaantanov @benjamingr


Thibaut:
Hey!

Thanks for the PR. I think we should include the guides, even if they're incomplete. 

Also:
- please make sure each entry has a `type` (in `EntriesFilter`). this will make browsing the docs via the sidebar easier.
- please don't edit `public/docs/docs.json`
- no need for `public/icons/docs/bluebird/logo.png`
- please squash

awalGarg:
Ok sure. I'll do it this weekend probably :)

Thibaut:
@awalGarg are you still working on this?

awalGarg:
@Thibaut thanks for the reminder! Yes, I am still interested in working on this. Will update this PR soon :)

Jeff-Lewis:
+1 :)

awalGarg:
I fixed everything but 1... help needed :((

> please make sure each entry has a type (in EntriesFilter). this will make browsing the docs via the sidebar easier.

@Thibaut I am not sure how to achieve this. For example: http://bluebirdjs.com/docs/api/catch.html doesn't specify what section it belongs to. It is visible only on the http://bluebirdjs.com/docs/api-reference.html page. Is there a feasible way I can carry forward that information from the reference page? (As in, technically ofcourse I can create a new state object and stuff, but does the scraper class provide any thing for that?)

Thibaut:
@awalGarg I don't see another way to achieve this. What you described sounds like the best solution.

You should be able to carry data across pages using the `context[:additional_options]` object (it's a hash so you can store anything on it, and it persists across pages, unlike the `context` object itself).

It'll be a bit of a hack but if it works I can clean up later (e.g. make it a proper `context[:env]` object).

awalGarg:
ah well, shall do soon. thanks!

jucrouzet:
Ping ?

Thibaut:
Closing due to inactivity. Feel free to open a new PR.

PR #292, opened 2015-11-14
_notif-close element is now an anchor instead of div

ESWAT:
Now closing a notification can be done with keyboard navigation and extensions like Vimium that look for anchor elements to act on.

Old behaviour is still retained when clicking on other anchors inside the notification or the notification itself.

### DIFF
--- a/assets/javascripts/views/misc/notif.coffee
+++ b/assets/javascripts/views/misc/notif.coffee
@@ -48,7 +48,7 @@ class app.views.Notif extends app.View
     return
 
   onClick: (event) =>
-    unless event.target.tagName is 'A'
+    if event.target.tagName isnt 'A' or event.target.classList.contains('_notif-close')
       $.stopEvent(event)
       @hide()
     return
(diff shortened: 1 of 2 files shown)


assets/javascripts/views/misc/notif.coffee
```diff
@@ -48,7 +48,7 @@ class app.views.Notif extends app.View
     return
 
   onClick: (event) =>
-    unless event.target.tagName is 'A'
+    unless event.target.tagName is 'A' && event.target.className != '_notif-close'
```
Thibaut: Could you change this to:
```coffee
if event.target.tagName isnt 'A' or event.target.classList.contains('_notif-close')
```

(and squash commits)

ESWAT:
Done :+1:

Thibaut:
Thanks! I'll merge and deploy this next week-end.

Thibaut:
Done: 097b328e7a0e8a29fcb17c27a4de93a23f10dcfd

PR #293, opened 2015-11-14
Add Dojo to devdocs

ShaneQful:
I also had to make a small change to the urlscraper to allow headers in the request, tests have been changed appropriately.


Thibaut:
Thanks for the PR. We need to find a way to automate the `self.initial_paths` list or this is going to be impossible to maintain.

One way to do it is to fetch the root page via Selenium. It'll run JS on the page, which may reveal all the links we need to follow. Angular works like that: https://github.com/Thibaut/devdocs/blob/master/lib/docs/scrapers/angular.rb#L33-L38 (to scrape with Selenium you'll need to run `SELENIUM=1 bundle install` and prefix the `thor` commands with `SELENIUM=1`). 

Another way to do it would be to override `root_page_body` (`include StubRootPage`) to fetch the `tree.json` file and turn it into an HTML list.

ShaneQful:
Fair enough, I'll look into both ways and resend the pull request with whichever seems to work best. 

Thanks for the tool and your time :)

Thibaut:
You could also update this PR. Whichever works best for you.

ShaneQful:
I've changed the code to override the root_page_body to get the tree.json file and turn it into HTML.

Thibaut:
Merged 59778f787711ee159dba0448f0792de13cf42642 346593354300588b5eaeb57155ec7b3c8370896d ebfe3a12080af85ae743ba73a84911c0830ca476 6939865137694565a45d6e83f21e8bf456f70a69
http://devdocs.io/dojo/

Thanks!

PR #337, opened 2016-01-23
Dual monitor sidebar stretch issue

chule143:
Using window.innerWidth instead of window.screenX for dual monitors


Thibaut:
Unfortunately this breaks the functionality for users not using dual monitors. Thanks for helping debug this issue, though. (a24a2ce1fe85f0938b05e951fe17ecc8bd2becb2 should fix it)

Issue #305, opened 2015-12-01
There are several accessibility issues with devdocs


ShaneQful:
There are several accessibility issues with devdocs which may impair users of assistive technologies such as screen reader users from using devdocs effectively:

* <s>Violations of WCAG colour contrast limits</s>
* There are several areas which could use aria roles could be used to improve the screenreader and keyboard usuability of the site
* Aria live regions would be good to allow screen readers know the documentation pane has updated 

I'd like to have a look at some of the issues above, just a heads up in case other people are working on the same

Thibaut:
Hey!

Good to know. There's definitely room for improvement.

> Violations of WCAG colour contrast limits

This is a tricky problem as I find that high contrast UIs don't look great on high quality displays (which already have good contrast). To get the best of both worlds, maybe we could create a separate high-contrast /  "accessibility" color scheme?

ShaneQful:
I'll probably fix the issues with it in high contrast mode first but the WCAG guidelines are supposed to be followed regardless of the mode. That said most of the application complies with the 4.5 minimum contrast.

ShaneQful:
@Thibaut is there a particular reason there is a meta tag which disables scaling, didn't want to just remove it as I presumed it was put there to solve a particular problem.

`<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=no,shrink-to-fit=no">`

Thibaut:
> is there a particular reason there is a meta tag which disables scaling?

The UI breaks when zoomed:

![iphone 5s - iphone 5s ios 9 2 13c75 2016-02-07 09-25-05](https://cloud.githubusercontent.com/assets/17579/12872877/2babef0c-cd7d-11e5-8dde-25aa98880a3e.png) ![iphone 5s - iphone 5s ios 9 2 13c75 2016-02-07 09-25-17](https://cloud.githubusercontent.com/assets/17579/12872879/2e9c7100-cd7d-11e5-8ee6-221ed250c618.png)


`user-scalable=no` also removes the 300ms tap/click delay on [Android](https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away) and soon [iOS](https://trac.webkit.org/changeset/191072).

ShaneQful:
It's technically a WCAG failure but considering all the desktop browsers ignore it for the most part and it's very useful for mobile there's no great point in removing it.

Thibaut:
[...]

PR #338, opened 2016-01-23
Make devdocs WCAG AA Colour Contrast Compliant

ShaneQful:
Obviously some of these changes are subjective but they are objective in making sure that devdocs is over the 4.5 WCAG AA colour contrast ratio in normal mode.

Pull request is related to but only fixes the contrast issues in normal mode mentioned in https://github.com/Thibaut/devdocs/issues/305

You can use the Axe plugin to verify the change:
https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd

### DIFF
--- a/assets/stylesheets/global/_variables.scss
+++ b/assets/stylesheets/global/_variables.scss
@@ -13,7 +13,7 @@ $documentBackground: #fafafa;
 
 $textColor: #333;
 $textColorLight: #666;
-$textColorLighter: #999;
+$textColorLighter: #707070;
 
 $inputFocusBorder: #35b5f4;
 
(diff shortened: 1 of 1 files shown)


Thibaut:
Thanks. I increased the contrast of links and "lighter text" in a different way in 6d9354f2127f9cae81f9b679a2eb54a9549bc8a8. "lighter text" isn't essential to use the app so IMO it's ok to keep it at a slightly low contrast. I didn't like the new "selection color", sorry; the current color almost meets the rule so I think we can keep it.

ShaneQful:
That's grand, that's why I mentioned that the changes were subjective. It's a lot easier to make sure an app is compliant if the colour palette is chosen to be so at the start of making an app. That way there doesn't have to be any compromises.

PR #348, opened 2016-02-01
Adding influxdb documentation

yosiat:
Hi,

I have added documentation to influxdb, I have some open questions/ideas:

1. I have scraped the documentation using UrlScrapper - but the documentation is open source - https://github.com/influxdata/docs.influxdata.com, I didn't know what is the preferred way for this url scrapping or file scrapping?
2. I removed the contribution guideline from everypage, which is something like this:
``
This documentation is open source. See a typo? Please, open an issue.
``
I think it will be nice to add to the attribution, what do you think?
3. It will be really easy to make this scrapper generic and support all InfluxData docs (for example Kapacitor, which I am interested in) - do you want me to do this?
4. The styling for h2 headers and etc, aren't applied, after I header "_influx" to _base.scss it's fixed, but I saw that the "new documentation guideline" and the commit log suggests that only @Thibaut only edits it..

That's it :)

BTW - I REALLY LOVE DEVDOCS, IT MAKES MY LIFE EASIER :+1:

### DIFF
--- a/lib/docs/filters/influxdb/clean_html.rb
+++ b/lib/docs/filters/influxdb/clean_html.rb
@@ -1,0 +1,18 @@
+module Docs
+  class Influxdb
+    class CleanHtmlFilter < Filter
+      def call
+        doc = @doc.at_css('#page-content')
+
+        # Re-position the page header
+        header = at_css('.page--body h1')
+        doc.children.first.add_next_sibling header
+
+        # Remove the contribution
+        at_css('.page--contribute').remove
+
+        doc 
+      end
+    end
+  end
+end
--- a/lib/docs/filters/influxdb/entries.rb
+++ b/lib/docs/filters/influxdb/entries.rb
@@ -1,0 +1,19 @@
+module Docs
+  class Influxdb
+    class EntriesFilter < Docs::EntriesFilter
+      
+      def get_name
+        at_css('#page-title h1').content
+      end
+
+      def get_type
+        # This is kinda hacky, we are fetching the current type from
+        # the url, we are asumming that the url pattern is
+        # category/page or category
+        path = current_url.relative_path_from(base_url)
+        path.split('/').first.titleize
+      end
+
+    end
+  end
+end
--- a/lib/docs/filters/influxdb/entries.rb
+++ b/lib/docs/filters/influxdb/entries.rb
@@ -11,7 +11,7 @@ module Docs
         # the url, we are asumming that the url pattern is
         # category/page or category
         path = current_url.relative_path_from(base_url)
-        path.split('/').first.titleize
+        "InfluxDB: #{path.split('/').first.titleize}"
       end
 
     end
--- a/lib/docs/scrapers/influxdb.rb
+++ b/lib/docs/scrapers/influxdb.rb
@@ -2,8 +2,8 @@ module Docs
   class Influxdb < UrlScraper
     self.name = 'InfluxDB'
     self.type = 'influxdb'
-    self.release = '0.9'
-    self.base_url = 'https://docs.influxdata.com/influxdb/v0.9/'
+    self.release = '0.10'
+    self.base_url = 'https://docs.influxdata.com/influxdb/v0.10/'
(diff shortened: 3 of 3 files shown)


Thibaut:
Hi

> I have scraped the documentation using UrlScrapper - but the documentation is open source - https://github.com/influxdata/docs.influxdata.com, I didn't know what is the preferred way for this url scrapping or file scrapping?

The way you did it is perfect. `FileScraper` is only needed for large documentations.

> I removed the contribution guideline from everypage

It's OK to leave it in. 

> It will be really easy to make this scrapper generic and support all InfluxData docs (for example Kapacitor, which I am interested in) - do you want me to do this?

Why don't we scrape all [4 docs](https://docs.influxdata.com/) under one doc/scraper? Seems they all go together + it'd make things easier to maintain & to toggle on/off in the app.

> The styling for h2 headers and etc, aren't applied

:+1: 

> BTW - I REALLY LOVE DEVDOCS, IT MAKES MY LIFE EASIER :+1: 

:heart: 

I have a few other PRs & updates to do before I can look into this one. ETA 3-4 weeks.

yosiat:
@Thibaut  you are suggesting to do something like this:

- InfluxData
 - InfluxDB
 - Kapacitor

If this is what you want, I don't agree with this because those are different products, you might use only InfluxDB + Telegraf and want to disable Kapacitor.

Thibaut:
>  you are suggesting to do something like this:

Yes, except we can still break up the categories (types) of each doc, instead of grouping all entries under "InfluxDB", "Kapacitor", etc.

Either that or we only add InfluxDB. The other projects are not very popular on GitHub and I'd rather not overload the UI with 3 separate docs — all owned by the same commercial entity.

yosiat:
Ok, How can implement this? how can I return parent type in EntriesFilter?

By the way, I believe that devdocs should add for each scrapper some kind of category, so we can distinct between "DB"/"Language"/"Frameworks"/etc so the ui won't be overloaded.

Thibaut:
> Ok, How can implement this? how can I return parent type in EntriesFilter?

Types can't be nested more than one level. But in `get_type` you can just return strings like `InfluxDB: Foo`, `InfluxDB: Bar`, `Kapacitor: Baz`, etc. so they'll be ordered by project in the sidebar.

> By the way, I believe that devdocs should add for each scrapper some kind of category, so we can distinct between "DB"/"Language"/"Frameworks"/etc so the ui won't be overloaded.

This is being tracked in #4. Unfortunately it's not as easy as it sounds and not on my short-term roadmap at the moment.

yosiat:
@Thibaut  do you wanted something like this: 
https://github.com/Thibaut/devdocs/commit/65f940534ebac4dee0575e0f28a83429e0e4ca9e

If so, I will fix the scrapper to handle all documentations (chrono, kapacitor, telegraf and influxdb).

But it will be nice to have some "type hierarchy" like this:
- InfluxData 
 - InfluxDB
   - Getting Started
    - Installing
 - Kapacitor
   - ...

Thibaut:
>  do you wanted something like this: 65f9405

Yes.

The Chef scraper is done in a similar way: http://devdocs.io/chef/ (Chef Client + Chef Server combined)

> But it will be nice to have some "type hierarchy" like this:

Agreed but IMO this would add too much complexity throughout the app, for a small benefit. I'd prefer keeping things simple, unless there's a clear benefit to a majority of users (or docs).

yosiat:
@Thibaut  what is the status of this pull request? are you waiting for me to implement something?

Thibaut:
@yosiat The process for getting this out isn't as simple as clicking the merge button. I'm working my way through https://github.com/Thibaut/devdocs/pulls and will look into yours as soon as I can. In the meantime please squash your commits.

Thibaut:
Merged: cd8590f90bdfc0613c2fcf9a1ec164b6c9df96b8 ebc6be321518695a2b215249051ec44a3bcf7dd4 
http://devdocs.io/influxdata/

PR #350, opened 2016-02-05
Add `ex` as alias for Elixir

AquarHEAD:
Because `.ex` is the extension used by Elixir's source files.


Thibaut:
Merged 43986630b0e4af2afadb188127a1aefff260d8f2
Thanks!

PR #353, opened 2016-02-09
Adding documentation for TensorFlow's API

Styrke:
Hey.

In spite of not having any experience with either ruby or css, I managed to add documentation for [TensorFlow](https://www.tensorflow.org/)'s python API.

Two notes:

1. Since TF has both a C++ and a python API, and probably will have more in the future, i set the release string to `0.6.0-py` to indicate that this is the python api. Let me know if you think something else would be better.
2. I can't figure out how to add syntax highlighting to the documentation. If I just try to make a copy of `sphinx.coffee` in `ssets/javascripts/views/pages/` and change line 3 to `class app.views.TensorflowPage extends app.views.BasePage` I can't load the DevDocs web app at all. Luckily syntax highlighting is not a deal breaker. :-)


Thibaut:
Thanks for the PR. Will look into this as soon as I can (ETA 3-4 weeks).

JonasMeinertz:
TensorFlow 0.7.0 was released since I created this PR, but I'm not sure about the preferred way to update the scraper.

Should I simply update lines [6 and 7 in `lib/docs/scrapers/tensorflow.rb`](https://github.com/Thibaut/devdocs/pull/353/files#diff-4a6ebf73f6aaeb2c1c47427d74fc4a14R6), or should I start keeping track of multiple versions of the documentation?

```ruby
    version '0.6.0-py' do
      self.release = '0.6.0-py'
      self.base_url = 'https://www.tensorflow.org/versions/0.6.0/api_docs/python/'
    end

    version '0.7.0-py' do
      self.release = '0.7.0-py'
      self.base_url = 'https://www.tensorflow.org/versions/r0.7/api_docs/python/'
    end
```

Thibaut:
I'd prefer we wait until 1.0 is released to keep track of multiple versions, so just change it to 0.7.0.

Would you mind also including the C++ docs? (change the base URL to https://www.tensorflow.org/versions/master/api_docs/)

JonasMeinertz:
@Thibaut Thanks for replying!

> I'd prefer we wait until 1.0 is released to keep track of multiple versions, so just change it to 0.7.0.

Okay. :+1: 

> Would you mind also including the C++ docs? (change the base URL to https://www.tensorflow.org/versions/master/api_docs/)

I would really prefer to keep them separate so C++ users don't get results for the python api and vice versa. That's why I included the `-py` identifier for the language in the release string, so the C++ api documentation can be e.g. `0.7.0-cc` or `0.7.0-cpp`. That way users will be able to pick exactly what they need, but the different languages would still be grouped together under the general *TensorFlow* item in the list of available apis.

I haven't included the C++ api yet because I don't use it myself, so it wouldn't get a lot of testing from my own side. But it should not be hard to try out.

What do you think about doing something like the this?:

```ruby
    version '0.7.0-py' do
      self.release = '0.7.0-py'
      self.base_url = 'https://www.tensorflow.org/versions/r0.7/api_docs/python/'
    end

    version '0.7.0-cc' do
      self.release = '0.7.0-cc'
      self.base_url = 'https://www.tensorflow.org/versions/r0.7/api_docs/cc/'
    end
```

Thibaut:
> What do you think about doing something like the this?:

Brilliant  :+1: 

I'd call the versions "Python" and "C++". The "+" in the version name might cause the app to :boom:. If so, change it to "Cpp" and I'll change it back / fix when I merge.

JonasMeinertz:
Just to be clear: Do you want me to call the versions "Python" and "C++" without the "0.7.0"-part?

Thibaut:
Correct

JonasMeinertz:
Great.

I'll get to it when I hear back about this problem with TensorFlow's C++ API documentation (tensorflow/tensorflow#1303) that I just ran into.

Thibaut:
Merged: cd55b861ca5001f2196137a517e71eb37adb0efc 2bec61a1a70cf532c1ddaf9bdd9b4c0a5d682a00
http://devdocs.io/tensorflow~python/ 
http://devdocs.io/tensorflow~cpp/ 

Thanks for the PR!

Styrke:
Awesome!

PR #355, opened 2016-02-09
Fix resizing left side list in Firefox

mikonieminen:
In Firefox drag events, different coordinates are always zeros. This
change will add window level dragover listener inside dragstart
handler of the resizer and removes the listener in dragend handler.
With this change resizing the list works for Firefox and for
Chrome.

### DIFF
--- a/assets/javascripts/views/layout/resizer.coffee
+++ b/assets/javascripts/views/layout/resizer.coffee
@@ -4,7 +4,6 @@ class app.views.Resizer extends app.View
   @events:
     dragstart: 'onDragStart'
     dragend: 'onDragEnd'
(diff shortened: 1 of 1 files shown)


mikonieminen:
I'm not too experienced with CoffeeScript and not sure if my approach of registering event handlers for window level is incorrect, but please advice if you prefer some other approach. To prevent dragover handler incorrectly interacting non-resizer drag events, I add the listener in dragstart and remove in dragend.

Thibaut:
Merged here: 1a04656d1165c3daad31e1cd35a844b581105fe5
Thanks for the fix :heart:

PR #356, opened 2016-02-11
Add Haxe documentation

JbIPS:
Hi,

I made a shot at creating an entry for Haxe documentation. It went rather smoothly except for the additional entries: since the language is organized on package and class, there is a lot of entries in each top level type. 

Is there a way to not show class fields in the nav bar but show them on search? Or maybe create multiple levels on the navigation? 

Improvement ideas are very welcomed :smile: !


Thibaut:
Thanks for the PR. Will look into this as soon as I can (ETA 3-4 weeks).

> Is there a way to not show class fields in the nav bar but show them on search?

Yes but I'd prefer all entries have a `type` / are accessible via the sidebar nav. If there are lots of types, we can group things together more aggressively.

For this PR I think we should group all the platform-specific entries together. I.e. a single type for "cpp", "java", etc. so we end up with the same top level types as on http://api.haxe.org/ 

> Or maybe create multiple levels on the navigation? 

This has come up a couple times but for now my position is that it would add too much complexity to the app, for a small payoff.

Thibaut:
Merged: daded63b6a9ea84c5188204c4fd75d3c106f5794 fbbf9d5d097b5ac9319e1d9da57dbb73160cc065 

Since there are so many entries, I ended up extracting the platform docs in separate docs:

http://devdocs.io/haxe/ 
http://devdocs.io/haxe~cpp/ 
http://devdocs.io/haxe~cs/ 
http://devdocs.io/haxe~flash/ 
http://devdocs.io/haxe~java/ 
http://devdocs.io/haxe~javascript/ 
http://devdocs.io/haxe~neko/ 
http://devdocs.io/haxe~php/ 
http://devdocs.io/haxe~python/

Thanks for the PR!

JbIPS:
Great idea to split it by target! Thank you :smile:

JesseTG:
If you're willing to do this with Haxe, why not do so with Java?

Thibaut:
@JesseTG not sure what you mean by "willing". I'm perfectly willing to add Java. The main reason it hasn't happened yet is that no one has open a PR for it, and getting it done would take hours (I do all this on my spare time).

PR #370, opened 2016-03-05
HTTPS links

pornel:
23 `s`es added


Thibaut:
Thanks but I'm not ready to make HTTPS the default just yet:

- some docs still have insecure assets (images)

- localStorage and IndexedDB are per-origin, so if a user switches from HTTP to HTTPS, they'll lose their offline data. To make HTTPS the default we'll need an upgrade path + to hard-redirect HTTP → HTTPS, so users don't accidentally go back and end up with outdated or missing offline data. 

I've been switching many docs/scrapers to HTTPS in recent months. Once 90%+ support HTTPS or don't have insecure assets, I'll make the switch.

This PR points to a few links that should be made protocol-relative, though. I'll fix that in a separate commit.

PR #382, opened 2016-04-06
Add Fortran documentation

scherepn:
The web documentation is available on the [GCC documentation site](https://gcc.gnu.org/onlinedocs/) as a tar.gz download.

Fortran doesn't have an "official" logo, so I just found something that was freely licensed and modified it to look good when shrunk to 16x16 and 32x32. It's available under the same terms as the original.

I included every chapter (1-10) because it was easy. If some sections don't make sense, I can easily remove them.


Thibaut:
Thanks. I'll take a look over the week-end.

Thibaut:
Merged: 56d9398148ee71fb018a5eec0ba0db8f3a795c65 6b37efda620225cc789feecbf274ad3577ba1cbd
http://devdocs.io/gnu_fortran/

Thanks for the PR!

Issue #341, opened 2016-01-25
feature request - collapse left-hand bar and add shortcut for popup search box


skibz:
sometimes, i find myself doing this if i only have one monitor at my disposal:

![screen shot 2016-01-25 at 13 09 01](https://cloud.githubusercontent.com/assets/2462211/12549269/d9224c08-c364-11e5-8381-ff8b685ea128.png)
(this is os x el capitan, using split-pane fullscreen workspaces)

the ability to hide the left-hand documentation list, making the main content full-width would allow me to further exploit my display's width. hiding the documentation list also hides the search box, so having a shortcut to bring one up or relocating it to the main content pane when the list is hidden would be needed.

thoughts?

stephanIOA:
+1

Thibaut:
@ccorcoveanu helped implement this feature. You can try it out by double-clicking the border of the sidebar. 

@skibz @stephanIOA any tweaks you'd like us to make?

skibz:
@ccorcoveanu @Thibaut thank you for adding this feature! it's purrrfect 🐱

luc-j-bourhis:
925217f is great on desktop but hard to use on mobile as the tap area is too small and there is no visual clue as to its location. Not on iOS at least.

Thibaut:
I tweaked the behavior of this a bit + made the feature discoverable via the new [preferences](http://devdocs.io/settings) page. Closing.

PR #384, opened 2016-04-11
Add double click event to toggle the sidebar

ccorcoveanu:
Linked to #341

### DIFF
--- a/assets/javascripts/app/app.coffee
+++ b/assets/javascripts/app/app.coffee
@@ -177,6 +177,14 @@
     # enabled because it doesn't support caching URLs that use CORS.
     @config[if @appCache and @settings.hasDocs() then 'index_path' else 'docs_host']
 
+  toggleSidebar: ->
+    state = @el.getAttribute('data-sidebar')
+    newState = if state == 'visible' then 'hidden' else 'visible'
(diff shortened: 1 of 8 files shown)


Thibaut:
Nice work!

Merged here: 4cb0d7a506c7a66634fc2f65a20257bb8a73acf3
Made changes here: 31495ff5a9ed51a2bed27adbc983d23d7c4c79df (merged the "sidebar" and "layout" settings to reduce duplication + tweaked the design)

As discussed via email, before announcing/finalizing the feature we need to:
- make the sidebar show/hide automatically when searching
- make the sidebar collapse/reappear when dragging the border all the way to the left or to the right

Let me know if I can help.

ccorcoveanu:
Thanks,

I'll commit the other changes in the next days.
Right now the dragging it's possible down to a minwidth, after that you have the forbidden cursor. Do you want it to stay this way and hide the side bar when the cursor reaches the left screen margin? Or should I remove that?

Thibaut:
> Do you want it to stay this way and hide the side bar when the cursor reaches the left screen margin?

Yes. I'd like to keep enforcing a minimum size on the sidebar, but if the user drags all the way to the left (the leftmost ~10 pixels), then the sidebar collapses.

PR #390, opened 2016-04-15
Add Pig Documentation

scherepn:
I did not include the [Pig logo](https://pig.apache.org/images/pig-logo.gif) because it would **not** look good shrunk to 16x16 and I couldn't find anything else that was logo-ish. I can try and find a generic cartoon pig that's freely licensed if it's absolutely required.

The documentation is all Apache Licensed.

https://pig.apache.org/

### DIFF
--- a/lib/docs/filters/pig/clean_html.rb
+++ b/lib/docs/filters/pig/clean_html.rb
@@ -1,0 +1,26 @@
+module Docs
+  class Pig
+    class CleanHtmlFilter < Filter
+      def call
+        @doc = at_css('#content')
+
+        css('.pdflink').remove
+
+        css('a[name]').each do |node|
+          node.next_element['id'] = node['name']
+        end
+
+        css('h2', 'h3').each do |node|
+          node.remove_attribute 'class'
+        end
+
+        css('table').each do |node|
+          node.remove_attribute 'cellspacing'
+          node.remove_attribute 'cellpadding'
+        end
+
+        doc
+      end
+    end
+  end
+end
--- a/lib/docs/filters/pig/entries.rb
+++ b/lib/docs/filters/pig/entries.rb
@@ -1,0 +1,33 @@
+module Docs
+  class Pig
+    class EntriesFilter < Docs::EntriesFilter
+
+      def include_default_entry?
+        false
+      end
+
+      def get_type
+        at_css('h1').content
+      end
+
+      def additional_entries
+        case slug
+          when 'basic', 'cmds', 'func'
+            nodes = css('h3')
+          when 'cont'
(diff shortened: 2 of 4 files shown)


Thibaut:
Didn't have time to do this one this week-end.

We'll need to find a logo for it (could always use the Apache logo). Otherwise the search results get confusing.

scherepn:
If it doesn't look too terrible, the [Hadoop logo](http://hadoop.apache.org/) would make more sense. If not, the apache logo is fine.

JesseTG:
Why not just the pig's head?

lib/docs/filters/pig/clean_html.rb
```diff
@@ -0,0 +1,26 @@
+module Docs
+  class Pig
+    class CleanHtmlFilter < Filter
+      def call
+        @doc = at_css('#content')
```
Thibaut: When you don't need to entire `<body>` to funnel through the other filters (e.g. to better identify the `type` in `EntriesFilter`), you can do this via `options[:container] = '#content'` in the scraper class.

lib/docs/filters/pig/entries.rb
```diff
@@ -0,0 +1,33 @@
+module Docs
+  class Pig
+    class EntriesFilter < Docs::EntriesFilter
+
+      def include_default_entry?
+        false
```
Thibaut: If the page has `additional_entries`, we should never return `false` here. Otherwise the URL without `#fragment` will 404.

I.e., if we have any entry for `/path#id`, we always need one for `/path`. I made that mistake many times in the past and it leads to broken links (within the doc, links are more likely to point to `/path` than `/path#id`).

lib/docs/filters/pig/entries.rb
```diff
@@ -0,0 +1,33 @@
...
+            nodes = css('h2, #diagnostic-ops + div > h3')
+          when 'perf'
+            nodes = css('h2, #optimization-rules + div > h3, #specialized-joins + div > h3')
+          else
+            nodes = css('h2')
+        end
+
+        nodes.each_with_object [] do |node, entries|
+          entries << [node.content, node['id'], get_type]
+        end
```
Thibaut: FYI this can also be written like this:
```ruby
nodes.map do |node|
  [node.content, node['id']]
end
```
The type already defaults to `get_type`.

Thibaut:
Just some comments for next time. No need to update. I'll get this out later today.

Thibaut:
Done: 1927f4e1960674b33177d361dcac681b3181f270 3b72e254c35190d546484c9659b560c275385d1d 
http://devdocs.io/apache_pig/

PR #391, opened 2016-04-15
Add Perl documentation

scherepn:
The Perl documentation is really messy and inconsistent, I did the best I could.

A download link can be found at the bottom of the [documentation page](http://perldoc.perl.org/). I tested this with the "Lite" version.

I wasn't sure what to put in the copyright section. The pages are freely licensed, but each one may be under different licenses and only some pages come with an explicit notice.

I used the onion logo because (despite being more widely used) the camel logo is owned by O’Reilly Media and can only be used when referring to their books, **not** Perl in general.


Thibaut:
Awesome work, thanks!

Merged here: c6da1d0c23d2d5c85781bdb7bac8e84728612f2b
Tweaked & finished here: f78b3658b291afdb06373e6c3403a12b9f611c89

http://devdocs.io/perl/

zicjin:
could anyone tell me the exact address of the document download link?

PR #393, opened 2016-04-17
Add NumPy documentation

Styrke:
I'm a little unsure about the license. It seems pretty custom, but Wikipedia claims that it's a "BSD-new license". The documentation itself claims that its "Copyright 2008-2009," but NumPy 1.10 was released in 2015 so I just made the attribution snippet say that.

The docs can be downloaded from http://docs.scipy.org/doc/. Look for the [HTML+zip] links.


Styrke:
I'm a little unsure about the license. It seems pretty custom, but Wikipedia claims that it's a "BSD-new license". The documentation itself claims that its "Copyright 2008-2009," but NumPy 1.10 was released in 2015 so I just updated it to say that.

The docs can be downloaded from http://docs.scipy.org/doc/. Look for the [HTML+zip] links.

Thibaut:
Thanks for the PR.

Seems the license is this: http://docs.scipy.org/doc/numpy/license.html

Would you mind confirming this with the NumPy develoeprs on their mailing list or bug tracker?

Styrke:
I'm not in doubt that that's the license. The question is what to call it in the (usually very short) attribution snippet. Perhaps it would be best to just provide a link like this:

```
&copy; Copyright 2008-2015, The Scipy community.<br>
Licensed under the <a href="http://docs.scipy.org/doc/numpy/license.html">Numpy License</a>.
```

Thibaut:
Merged: 547f4f54edb03290a81457c3a3ebcd6e48c69cda 6d36b339e023601a5085d081300e319989452661 
http://devdocs.io/numpy/

Thanks for the PR!

PS: I looked into adding SciPy as well, but the problem is that the doc requires a 60KB JS library (MathJax) to display all the math symbols & equations :/

Styrke:
> PS: I looked into adding SciPy as well, but the problem is that the doc requires a 60KB JS library (MathJax) to display all the math symbols & equations :/

Perhaps the LaTeX formulas can be converted to images while scraping the documentation. For example I think Wikipedia generates PNGs serverside.

I found [this ruby gem](https://rubygems.org/gems/latex2image/versions/0.1.0) to do it, but it will require a full install of TeX Live to do the conversion while scraping.

Thibaut:
PNGs won't do because the scrapers only support static HTML files (and PNGs don't work offline). We could try https://github.com/gjtorikian/mathematical to convert to inline SVGs, but that also likely comes with heavy dependencies (though if it works reliably and isn't too slow then I don't mind it).

PR #394, opened 2016-04-24
Hide/Show sidebar when dragging it or typing into the serarch input

ccorcoveanu:
Added the new interactions that could trigger hiding/showing of the sidebar.
If you drag it all the way to the edge it will hide, then if you drag it to the right it will show again.
If you deliberately hide the sidebar then type into the search input, the sidebar will be shown. If you delete the text, the sidebar will be hidden again. This behavior does not apply if the sidebar is already opened.
If I missed something or the behavior is not what you were expecting let me know and I will change it. 
It may be that you'll want me to refactor one or two things, just let me know.

This additions are related to #341

### DIFF
--- a/assets/javascripts/app/settings.coffee
+++ b/assets/javascripts/app/settings.coffee
@@ -79,6 +79,15 @@ class app.Settings
     catch
     return
 
+  hasLayout: (name) ->
+    try
+      layout = (Cookies.get(LAYOUT_KEY) || '').split(' ')
+      $.arrayDelete(layout, '')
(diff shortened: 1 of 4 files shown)


Thibaut:
Why did you close?

ccorcoveanu:
I was in a hurry and forgot a console log. I'll clean it and resubmit the pull request.

Thibaut:
Ok. You didn't have to close the PR for that (can just push a new commit, or amend your original commit and force push), but no worries :)

PR #398, opened 2016-05-01
Hide/Show sidebar when dragging it or typing into the serarch input

ccorcoveanu:
Added the new interactions that could trigger hiding/showing of the sidebar.
If you drag it all the way to the edge it will hide, then if you drag it to the right it will show again.
If you deliberately hide the sidebar then type into the search input, the sidebar will be shown. If you delete the text, the sidebar will be hidden again. This behavior does not apply if the sidebar is already opened.
If I missed something or the behavior is not what you were expecting let me know and I will change it. 
It may be that you'll want me to refactor one or two things, just let me know.

This additions are related to #341

### DIFF
--- a/assets/javascripts/views/layout/document.coffee
+++ b/assets/javascripts/views/layout/document.coffee
@@ -46,6 +46,27 @@ class app.views.Document extends app.View
     app.appCache?.updateInBackground()
     return
 
+  hideSidebar: (saveLayout = true) ->
+    sidebarHidden = app.el.classList.contains(HIDE_SIDEBAR_CLASS)
+    return if sidebarHidden
+    app.el.classList.add(HIDE_SIDEBAR_CLASS)
+    return unless saveLayout
+    app.settings.setLayout(HIDE_SIDEBAR_CLASS, true)
+    app.appCache?.updateInBackground()
+    return
+
+  showSidebar: (saveLayout = true) ->
+    sidebarHidden = app.el.classList.contains(HIDE_SIDEBAR_CLASS)
+    return unless sidebarHidden
+    app.el.classList.remove(HIDE_SIDEBAR_CLASS)
+    return unless saveLayout
+    app.settings.setLayout(HIDE_SIDEBAR_CLASS, false)
+    app.appCache?.updateInBackground()
+    return
+
+  hasSidebar: ->
+    return !app.el.classList.contains(HIDE_SIDEBAR_CLASS) && !app.settings.hasLayout(HIDE_SIDEBAR_CLASS)
+
   setTitle: (title) ->
     @el.title = if title then "DevDocs - #{title}" else 'DevDocs API Documentation'
 
--- a/assets/javascripts/views/layout/document.coffee
+++ b/assets/javascripts/views/layout/document.coffee
@@ -39,28 +39,12 @@ class app.views.Document extends app.View
     app.appCache?.updateInBackground()
     return
 
-  toggleSidebar: ->
-    sidebarHidden = app.el.classList.contains(HIDE_SIDEBAR_CLASS)
-    app.el.classList[if sidebarHidden then 'remove' else 'add'](HIDE_SIDEBAR_CLASS)
-    app.settings.setLayout(HIDE_SIDEBAR_CLASS, !sidebarHidden)
-    app.appCache?.updateInBackground()
-    return
-
-  hideSidebar: (saveLayout = true) ->
-    sidebarHidden = app.el.classList.contains(HIDE_SIDEBAR_CLASS)
-    return if sidebarHidden
-    app.el.classList.add(HIDE_SIDEBAR_CLASS)
+  toggleSidebar: (saveLayout = true) ->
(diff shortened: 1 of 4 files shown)


assets/javascripts/views/layout/document.coffee
```diff
@@ -46,6 +46,27 @@ class app.views.Document extends app.View
     app.appCache?.updateInBackground()
     return
 
+  hideSidebar: (saveLayout = true) ->
```
Thibaut: Rather than having extra `hideSidebar` / `showSidebar` functions, I'd prefer we use and extend the existing `toggleSidebar` function (maybe with a boolean argument).

assets/javascripts/views/layout/document.coffee
```diff
@@ -46,6 +46,27 @@ class app.views.Document extends app.View
...
+    sidebarHidden = app.el.classList.contains(HIDE_SIDEBAR_CLASS)
+    return unless sidebarHidden
+    app.el.classList.remove(HIDE_SIDEBAR_CLASS)
+    return unless saveLayout
+    app.settings.setLayout(HIDE_SIDEBAR_CLASS, false)
+    app.appCache?.updateInBackground()
+    return
+
+  hasSidebar: ->
+    return !app.el.classList.contains(HIDE_SIDEBAR_CLASS) && !app.settings.hasLayout(HIDE_SIDEBAR_CLASS)
```
Thibaut: Why do we need the extra `!app.settings.hasLayout(HIDE_SIDEBAR_CLASS)` check?

assets/javascripts/views/search/search.coffee
```diff
@@ -68,8 +68,10 @@ class app.views.Search extends app.View
     @value = @input.value
 
     if @value.length
+      app.document.showSidebar(false)
```
Thibaut: This would be better done in the `app.views.Sidebar` class, which already toggles the search result view on/off:
[sidebar.coffee#L16-L18](https://github.com/Thibaut/devdocs/blob/4468f65dd86bead162ab5f4a39d1b853a0c55214/assets/javascripts/views/sidebar/sidebar.coffee#L16-L18)
[sidebar.coffee#L49-L55](https://github.com/Thibaut/devdocs/blob/4468f65dd86bead162ab5f4a39d1b853a0c55214/assets/javascripts/views/sidebar/sidebar.coffee#L49-L55)

Thibaut:
Thanks for the PR and sorry for the delay.
Let me know when you've pushed a new version or if you'd prefer I make the changes.

assets/javascripts/views/layout/document.coffee
ccorcoveanu: Here we're checking for both class name and cookie value. They could be different in the case when the user hides the sidebar then types in the search field.

assets/javascripts/views/layout/document.coffee
ccorcoveanu: I'll see how I can do this without making the `toggleSidebar` function too messy

ccorcoveanu:
I'll make the changes today, thanks for the notes :)

assets/javascripts/views/search/search.coffee
ccorcoveanu: This would be an issue because the `showDocList` is triggered also on page load. We would then be in the situation where we would want 2 different results calling `toggleSidebar` from the same state.
1. Sidebar hidden - page loads -> call to `showDocList`-> call to `toggleSidebar` (here we would want to keep it hidden)
2. Sidebar hidden - user types into the searchbox -> call to `toggleSidebar` (here we would want to show the sidebar)

The only way we can difference this two actions using the `onInput` function. Or maybe am I missing something

ccorcoveanu:
I pushed another commit removing the hide/show methods. I didn't add another boolean inside the `toggleSidebar` method but it came with the cost of having this extra check:
`forceShow = (!hasHiddenClass || !@hasSidebar()) && !saveLayout`
Also, for the remaining part, of moving the logic of hiding/showing of the sidebar from `app.views.Search` to `app.views.Sidebar` I will wait for your input :)

Thibaut:
Merged here: dd4b9bfa1ffb96552e83bc49f4ba08df99e60ff7
Made some changes here: 4fc112601588a44c8245542fe1e08b8922bed552
And eventually decided that hiding the sidebar by resizing isn't worth the extra code complexity + the UX doesn't feel great: 740bcc65ba8176d17aa096a7ec53f3bffe954c45

Thanks for your help with this feature and sorry for removing that last bit. I'll look into announcing the feature in the next week or two and let you know if there's anything left to do.

ccorcoveanu:
Thanks!
I think there's a small bug. If you hide the sidebar, then refresh the page, it will toggle to visible. You will then get the tooltip that the website was updated and on the next refresh will be in the correct state.
I can reproduce this everytime.

Thibaut:
Good catch. I forgot to re-add `app.appCache?.updateInBackground()` in `toggleSidebar`. Will fix later.

PR #406, opened 2016-05-14
let heading wrap

spacewander:
So that we can show longer topics like http://devdocs.io/redis/zunionstore


Thibaut:
This change breaks the design. I updated the Redis doc with shorter headings instead.

PR #407, opened 2016-05-16
Fix URL#relative_path_to

### LINKED ISSUE #402: Local file names from slugs are not being sanitized
*Repeatable:* When remote slugs contain characters unsafe for the local file system (in my case NTFS on Windows 7)

*Steps to repeat:*
1. Clone my [fork](https://github.com/nucular/devdocs/tree/love) for the LÖVE scraper (at the time of writing HEAD of the branch `love`)
2. Run `thor docs:page love Canvas:renderTo` (which might fail because the `docs:page` command is broken, just remove the check at [lib/tasks/docs.thor#L34:L36](https://github.com/Thibaut/devdocs/blob/master/lib/tasks/docs.thor#L34:L36) for now).

*Expected result:* `public/docs/love/canvas:renderto.html` (or similar) is generated.

*Actual result:* `public/docs/love/canvas` is generated.

While writing a scraper for the [LÖVE](http://love2d.org) engine documentation I ran into an issue where remote pages with colons in their slugs (for example [Canvas:renderTo](https://love2d.org/wiki/Canvas:renderTo)) are being written to files named like the slug up to the first colon (for example `public/docs/love/canvas`). This results both in subsequent pages overwriting this file and in the client not being able to find this file under the expected path.

I would fix this problem myself in a pull-request, but I am neither sure where path sanitization logic would be appropriately placed (I assume in `Docs.Doc#path`) nor what to replace unsafe characters with to keep the integrity of other documentations.

nucular:
Replace File.basename in URL#relative_path_to because it doesn't handle special characters in URLs well.
Since this branch will only be executed when `dest` does not end with `'/'`, there is no need to handle this special case, so a simple `dest.split('/').last` should suffice.
Fixes #402 completely, together with 2813cf9be32cf6f2dc69abca75069084d871aeb6.

### DIFF
--- a/lib/docs/core/url.rb
+++ b/lib/docs/core/url.rb
@@ -100,7 +100,7 @@ module Docs
           result << '/' if result != '.'
         end
       else
-        dest_dir.parent.relative_path_from(base_dir).join(::File.basename(dest)).to_s
+        dest_dir.parent.relative_path_from(base_dir).join(dest.split('/').last).to_s
       end
     end
 
--- a/test/lib/docs/core/url_test.rb
+++ b/test/lib/docs/core/url_test.rb
@@ -394,6 +394,14 @@ class DocsUrlTest < MiniTest::Spec
         assert_equal 'file', url.relative_path_to('http://example.com/file?query#frag')
       end
 
+      it "returns 'some:file' with 'http://example.com/some:file'" do
+        assert_equal 'some:file', url.relative_path_to('http://example.com/some:file')
+      end
+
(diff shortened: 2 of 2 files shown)


Thibaut:
Could you add a [regression test](https://github.com/Thibaut/devdocs/blob/1db5a857ddbf5a0ca2c72a1cedf3f5b5ccf7e3ec/test/lib/docs/core/url_test.rb#L279-L423) please?

nucular:
Are these two tests sufficient?

Thibaut:
Merged: 034ecfae72f301519c570bd14668ffff8bf409ce
Thanks

PR #415, opened 2016-06-02
Add vim-www to the list of plugins

waiting-for-dev:
Even if it is not a dedicated DevDocs plugin, it integrates it between its
defaults. So, OOTB you can do:

    :Wsearch devdocs some_method

While, adding the following to `~/.vimrc`:

    `let g:www_shortcut_engines = { 'devdocs': ['DevDocs'] }`

Provides with:

    :DevDocs some_method


Thibaut:
Hey

Since this is more a generic search plugin than a 1-click integration with DevDocs like the other tools in the list, I think a link on the wiki like you've already added is enough. I don't want the in-app list to get too long.

Thanks for integrating DevDocs in your plugin though!

PS: I just retweeted [your tweet](https://twitter.com/waiting_for_dev/status/738435848488820736)

waiting-for-dev:
OK, no problem :) I thought that it would be interesting to have a link to work with DevDocs in a major editor like vim, at least until there is something completely specific. But your argument is totally reasonable. Thanks for your job in DevDocs! :)

PR #416, opened 2016-06-02
Add API doc for Padrino

wikimatze:
I have already discussed this issue on the mailing list (https://groups.google.com/forum/#!topic/devdocs/mAMiT-LZu14).
If everything is working fine, I'll add the other options for Padrino.

Cheers,
Matthias

### DIFF
--- a/lib/docs/filters/padrino/clean_html.rb
+++ b/lib/docs/filters/padrino/clean_html.rb
@@ -2,6 +2,8 @@ module Docs
   class Padrino
     class CleanHtmlFilter < Filter
       def call
+        css('.summary_toggle').remove
+        css('.inheritanceTree').remove
         at_css('#content')
       end
     end
--- a/lib/docs/filters/padrino/clean_html.rb
+++ b/lib/docs/filters/padrino/clean_html.rb
@@ -1,0 +1,18 @@
+module Docs
+  class Padrino
+    class CleanHtmlFilter < Filter
+      def call
+        title = at_css('#filecontents h1')
+        @doc = at_css('#content')
+
+        css('.highlight').each do |node|
+          node.name = 'pre'
+          node['data-lang'] = node.at_css('[data-lang]')['data-lang']
+          node.content = node.content
+        end
+
+        doc
+      end
+    end
+  end
+end
--- a/lib/docs/filters/padrino/clean_html.rb
+++ b/lib/docs/filters/padrino/clean_html.rb
@@ -2,16 +2,7 @@ module Docs
   class Padrino
     class CleanHtmlFilter < Filter
       def call
-        title = at_css('#filecontents h1')
-        @doc = at_css('#content')
-
-        css('.highlight').each do |node|
-          node.name = 'pre'
-          node['data-lang'] = node.at_css('[data-lang]')['data-lang']
-          node.content = node.content
-        end
-
-        doc
+        at_css('#content')
       end
     end
   end
--- a/lib/docs/filters/padrino/entries.rb
+++ b/lib/docs/filters/padrino/entries.rb
@@ -2,7 +2,9 @@ module Docs
   class Padrino
     class EntriesFilter < Docs::EntriesFilter
       def get_name
-        at_css('h1, h2').content
+        name = at_css('h1, h2').content
+        name.remove! 'Class: '
+        name.remove! 'Module: '
       end
 
       def get_type
--- a/lib/docs/filters/padrino/entries.rb
+++ b/lib/docs/filters/padrino/entries.rb
@@ -1,0 +1,17 @@
+module Docs
+  class Padrino
+    class EntriesFilter < Docs::EntriesFilter
+      def get_name
+        node = at_css('h1')
+        result = node.content.strip
+        result << ' event' if type == 'Events'
+        result << '()' if node['class'].try(:include?, 'function')
+        result
+      end
+
+      def get_type
+        return 'Ruby'
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


lib/docs/filters/padrino/entries.rb
```diff
@@ -0,0 +1,17 @@
...
+      def get_name
+        node = at_css('h1')
+        result = node.content.strip
+        result << ' event' if type == 'Events'
+        result << '()' if node['class'].try(:include?, 'function')
+        result
+      end
+
+      def get_type
+        return 'Ruby'
```
Thibaut: We need to group entries by class here ^

lib/docs/scrapers/padrino.rb
```diff
@@ -0,0 +1,20 @@
+module Docs
+  class Padrino < UrlScraper
+    self.name = 'padrino'
+    self.slug = 'padrino'
+    self.type = 'ruby'
+    self.version = 'master'
```
Thibaut: This should be `self.release = ` now, with the actual version number.

lib/docs/filters/padrino/clean_html.rb
```diff
@@ -0,0 +1,18 @@
+module Docs
+  class Padrino
+    class CleanHtmlFilter < Filter
+      def call
+        title = at_css('#filecontents h1')
```
Thibaut: This variable isn't used ^

lib/docs/filters/padrino/clean_html.rb
```diff
@@ -0,0 +1,18 @@
...
+        @doc = at_css('#content')
+
+        css('.highlight').each do |node|
```
Thibaut: What page is this for? I can't seem to find any `.highlight` element on the website.

lib/docs/filters/padrino/entries.rb
```diff
@@ -0,0 +1,17 @@
+module Docs
+  class Padrino
+    class EntriesFilter < Docs::EntriesFilter
+      def get_name
+        node = at_css('h1')
+        result = node.content.strip
+        result << ' event' if type == 'Events'
+        result << '()' if node['class'].try(:include?, 'function')
```
Thibaut: AFAICT the two lines above are unnecessary / not doing anything.

Thibaut:
Thanks for the PR. This needs a bit more work before I can get it out. See inline comments.

lib/docs/filters/padrino/entries.rb
wikimatze: What do you mean by this?

Thibaut: The type is used to categorize entries in the sidebar:

devdocs - nokogiri documentation 2016-06-04 12-21-52

We need to return different ones based on the class, module, etc.  For example:

https://github.com/Thibaut/devdocs/blob/ebd6a2204da031328c1ba707e9d119e3969b4146/lib/docs/filters/nokogiri2/entries.rb#L4-L8

wikimatze: Makes sense, the generated HTML looks now more like that.

Thibaut: Cool. Could you post a screenshot of what it looks like in the app please?

wikimatze: I mean the output in the terminal when generating the HTML

generated_html_output


The navigation in the HTML is still missing (think some JS is missing there)

generated_html

Thibaut: Ok. We need to remove these `Class: ` in the entries' names. Also we should add entries (using `additional_entries`) for all the methods. Right now it looks like it's only indexing modules/classes. See [here](https://github.com/Thibaut/devdocs/blob/ebd6a2204da031328c1ba707e9d119e3969b4146/lib/docs/filters/rdoc/entries.rb#L30-L48) for an example. Documentation is [here](https://github.com/Thibaut/devdocs/wiki/Filter-Reference#entriesfilter).

> The navigation in the HTML is still missing (think some JS is missing there)

You're not supposed to open the HTML files directly. The app doesn't work like that. Run `rackup` in the root directory. That will boot the server. Then you can access the app at http://localhost:9292/

wikimatze: Ahh, didn't know about the `rackup` commands - it's already looking awesome, I totally love it. The `Class:` and `Module:` has been removed:
no_class_module

wikimatze: Hi @Thibaut methods and classes are now indexed (I checked some links in the original one and the generated one and I got same results). Now it's time to make the highlighting!

Thibaut:
Merged: http://devdocs.io/padrino/ e1e4a626ffeac8febb18cce32ed9015e6fb9d2e4 481233050daced191d7a89c022d64988212f0518

I didn't have time to make the source code blocks work properly (hidden by default + link to show/hide), so I removed them.

Would be cool to get the [guides](http://padrinorb.com/guides/) in there as well, but DevDocs doesn't support scraping multiple origins within a single scraper at the moment. Maybe later.

Thanks again for the PR & updating!

wikimatze:
No problems with the code block, maybe I can work on them, when there is the need someone misses them.

Yeah having the guides in there would be awesome. Why not writing a second scraper which will be executed after the first one (let's name is `guides_scraper`).

Thibaut:
> Yeah having the guides in there would be awesome. Why not writing a second scraper which will be executed after the first one (let's name is `guides_scraper`).

This would add a second entry in the list of docs for Padrino. That list is already pretty long + I'd rather keep the data clean, with each doc encapsulated in a single package.

Rails and a few other docs have the same problem/need, so I think I'll take a stab at adding multi-origin support sometime soon.

wikimatze:
Okay, what having a second scrape (says 'Padrino docs scraper' and 'Rails docs scraper')?

PR #447, opened 2016-07-26
Add Docker documentation

scherepn:
The docker docs includes both their Free/Open Source Software tools as well as their proprietary ones. This scraper only includes the FOSS ones.

### DIFF
--- a/lib/docs/scrapers/docker.rb
+++ b/lib/docs/scrapers/docker.rb
@@ -1,0 +1,42 @@
+module Docs
+  class Docker < UrlScraper
+    self.name = 'Docker'
+    self.slug = 'docker'
+    self.type = 'docker'
+
+    self.links = {
+      home: 'https://docker.com/',
+      code: 'https://github.com/docker/docker'
+    }
+
+    html_filters.push 'docker/clean_html', 'docker/entries'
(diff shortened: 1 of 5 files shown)


lib/docs/scrapers/docker.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    options[:only_patterns] = [
+      /\Aengine\//,
+      /\Acompose\//,
+      /\Amachine\//,
+      /\Aswarm\//
+    ]
+
+    options[:skip] = [
+      "engine/installation/ubuntulinux/" # invalid document
```
scherepn: I think this is a bug in the document parser. Instead of responding with a 301/302, docker.com instead responds with a "valid" web page that triggers a redirect via an HTML header. This line is just here so the scraper doesn't crash.

Thibaut:
Thanks for the PR. I'll try to get this out in the next couple weeks.

Thibaut:
Done: 457439f35a1e3d96a23d008c47dc0b6f3ab397c1 e92d0e15e79e9bc8ef41ae28eedd4a0c8859dfbd
http://devdocs.io/docker/

PR #473, opened 2016-08-30
Autofocus search input when opening page

th-we:
For some reason, on my Firefox/Linux (even when started with `-safe-mode` and/or `-private`), the search field is not focused. This pull request should fix that.

As far as I can see, `<input autofocus />` does not cause the keyboard to pop up on touch devices, which would probably be undesired.

### DIFF
--- a/views/app.erb
+++ b/views/app.erb
@@ -5,7 +5,7 @@
     <button type="button" class="_mobile-btn _menu-btn">Menu</button>
     <button type="button" class="_mobile-btn _home-btn">Home</button>
     <form class="_search" role="search">
-      <input type="search" class="_search-input" placeholder="Search&hellip;" autocomplete="off" autocapitalize="off" autocorrect="off" spellcheck="false" maxlength="30" aria-label="Search">
+      <input type="search" class="_search-input" placeholder="Search&hellip;" autocomplete="off" autocapitalize="off" autocorrect="off" spellcheck="false" maxlength="30" aria-label="Search" autofocus="autofocus">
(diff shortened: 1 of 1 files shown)


th-we:
Thinking about it, the real issue might be why [this line of code](https://github.com/Thibaut/devdocs/blob/6f16945418c4f0d2fffba40096cdc33fb6aa27b5/assets/javascripts/views/searc/search.coffee#L45) doesn't seem to do its job for Firefox/Linux (v. 48.0). Windows works (48.0.2).

Thibaut:
Merged: 14e85a641e0af3e417ea1687b873a4b3ac2f5120
Hopefully this fixes the issue.

Thibaut:
I had to revert this because it causes the keyboard to pop up on iOS when the app is added to the homescreen. The actual bug sounds like a browser bug, though. What does `$.isTouchScreen()` in the web console return on Linux?

th-we:
`$.isTouchScreen()` returns `true` (on Ubuntu 16.04 - I checked both under LXDE and Unity desktops). That's obviously because `ontouchstart` etc. do exist.  I'm not sure whether this depends on how Firefox is built.  I checked the configure options that were used to build Firefox and saw nothing suspicious.  There's no touch screen attached to this laptop. There is a [Stackoverflow comment](http://stackoverflow.com/questions/3974827/detecting-touch-screen-devices-with-javascript#comment-19047102) on this topic, that however doesn't help much.

Thibaut:
@th-we no worries. `$.isTouchScreen()` could probably be improved to detect touch devices that also have a keyboard.

Issue #331, opened 2016-01-18
MathML source passed through


echristopherson:
On pages with MathML, devdocs seems to pass through the MathML source instead of rendering it. For instance, <http://devdocs.io/javascript/global_objects/math/pi> shows `\mathtt{\mi{Math.PI}} = \pi \approx 3.14159`.

echristopherson:
I see I ran up against Chrome's known lack of support for MathML. When I go to <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/PI>, the page tells me "Your browser does not support MathML. A CSS fallback has been used instead.", and it does use a fallback which displays correctly. I wonder if DevDocs could employ a similar fallback.

japborst:
I think [MathJax](https://www.mathjax.org/) could be very well employed for this. Also see [MathML input](http://docs.mathjax.org/en/latest/start.html#mathml-input) in the docs.

Thibaut:
IMO MathJax is too big/slow for DevDocs. @japborst added a CSS fallback in https://github.com/Thibaut/devdocs/pull/496 which I think is enough to close this issue.

PR #496, opened 2016-09-26
Add MathML CSS fallback

japborst:
As indicated in e.g. #331, MathML is not supported in DevDocs in unsupported browsers such as Chrome.  This PR uses the same CSS fallback [Mozilla uses](https://developer.cdn.mozilla.net/static/styles/libs/mathml.css) in their docs, that originates from this repo [fred-wang/mathml.css](https://github.com/fred-wang/mathml.css): example [Math.PI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/PI) This method is doc-agnostic, so will also work for all other docs that uses MathML. 

Process is as follows:
 1. Check for MathML support
 2. Sideload CSS if not existent

This method saves bandwidth and is thus faster if MathML is supported. However, currently it sideloads the CSS externally.

2 possible alternatives:
 1. Sideload the CSS locally (couldn't figure out how to do that in the set-up of DevDocs, unfortunatly)
 2. Alter the css with a `.fallback` class, include in `application.css.scss` and apply class if not supported. However, the docs are not present on load so applying the class makes it difficult.

Whilst [MathJax](https://www.mathjax.org/) provides much better support, it is rather immense...

Feedback is welcome.

### DIFF
--- a/assets/javascripts/vendor/mathml.js
+++ b/assets/javascripts/vendor/mathml.js
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. 
+ * Adapted from: https://github.com/fred-wang/mathml.css */
+/*jslint browser: true*/
+
+(function () {
+    "use strict";
+    window.addEventListener("load", function () {
+        var box, div, link, namespaceURI;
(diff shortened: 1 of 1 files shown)


Thibaut:
Merged in 4c28cb6dbdd22b58cee98d1824c0f1f59d45d97d and added the local CSS in a8bd4227c3377d63b7e018ecb95f2eaff9c7328d. Thanks!

PR #499, opened 2016-10-01
Remove different pandas versioning

japborst:
Pandas shows a dropdown as if there were multiple versions to be maintained, but there is only one version here.

### DIFF
--- a/lib/docs/scrapers/pandas.rb
+++ b/lib/docs/scrapers/pandas.rb
@@ -2,6 +2,8 @@ module Docs
   class Pandas < UrlScraper
     self.name = 'pandas'
     self.type = 'sphinx'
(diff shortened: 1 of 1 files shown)


Thibaut:
AFAICT the doc is versioned, so we can do the same in DevDocs:
http://pandas.pydata.org/pandas-docs/version/0.19.0/
http://pandas.pydata.org/pandas-docs/version/0.18.1/
Given that it's a 0.x release we should probably not have versioned it to begin with, but now that it's done, I think it's best to stick with it.

japborst:
Fair enough, I understand your reasoning.

They have a little odd version increment policy, though. About twice a year a new 0.x is released. API changes are never big and docs always list deprecation or different behaviour across versions.

PR #500, opened 2016-10-02
iPython+notebook syntax highlighting and allowing data:image uris

japborst:
In some python-based documentation, jupyter notebook or iPython code cells are used. These are not properly recognised by the sphinx html cleaner. Example: http://devdocs.io/pandas~0.18/10min

Additionally, such notebooks may contain data:image URIs for images, which are currently replaced by '#' as those are deemed invalid. This fix simply skips any sanitising for these.


Thibaut:
Thanks! Merged 8e2e150be11df55e0571491cb670775c65001c6d and added regression test in 721adf8e211f431f410950189447f4924c4fd5a9.

PR #539, opened 2016-11-30
Codeception documentation

edno:
* Name: Codeception
* Source: https://github.com/Codeception/Codeception
* Docs: http://codeception.com/docs/index.html
* Docs license: https://github.com/Codeception/Codeception/blob/master/LICENSE


Thibaut:
Thanks for the PR.

Merged here: 72ab3435725d1236123bba6678bd91cbb1c3f8fe
Finished here: 802a8d38046cb299a59da0858fa15ce61444e925
http://devdocs.io/codeception/

Some notes if you've got plans to contribute more docs:
- I switched the syntax highlighting code to be more generic a while ago (sorry the wiki/docs aren't great :/). See [`simple.coffee`](https://github.com/Thibaut/devdocs/blob/master/assets/javascripts/views/pages/simple.coffee). Just add a `data-language` attribute (and remove all existing highlighting) to the `pre` elements in the `clean_html` filter, then no need for a dedicated coffee file. 
- For readability, all pages should begin with a top-level `<h1>` element.
- It's sometimes easier to run the `entries` filter before `clean_html`, or vice-versa (e.g. when `clean_html` modifies headings).
- `include_default_entry?` should never return `false` when the page has additional entries. I did this in the past and it causes issues, because then `/page#frag` works, but `/page` doesn't, which leads to broken links. The only remaining use-case for `include_default_entry?` is when you want to visit a page but not index/store it at all (e.g. if the page has links not found on other pages).
- `self.release` → full release number
- Indent with 2 spaces.

Lastly: I may not always accept docs with less than ~3k GitHub stars, as it takes time to maintain/update them. Feel free to email me to check before doing the work.

edno:
Thanks @Thibaut for the notes, I will use them for the next scraper. Next documentation will be [Composer](https://github.com/composer/composer) with more than 3k start.

PR #553, opened 2016-12-15
[WIP] First exportable version of sequel

wmaciejak:
For quick export and generate sequel devdocs you need to execute
thor docs:generate sequel --force --debug
Docs don't look so nice, but the main part of exporting is done. Now it’s the time for analyze rdoc source code and rewrite actual filters.


truepalyer6:
W

PR #556, opened 2016-12-18
Rustdoc scraper: add rule to remove "Run" links

ulidtko:
All over the Rust docs, there's this problem:

screenshot

It obviously comes from the live-test "Run" button in the original docs:

screenshot

Hence, this commit filters out the `test-arrow` links to fix the problem.


Thibaut:
Thanks for the PR! I ended up fixing this in 4bf382cad15c835eacaab3fa1f48f9ac0d7e3381 before looking at your PR.

PR #560, opened 2016-12-27
Java 8 scraper

japborst:
* Used Oracle as the [source](http://docs.oracle.com/javase/8/docs/api/)
* Included the following. Use those ones the most myself and per suggestions on [Trello](https://trello.com/c/o5uswXmX/208-java-openjdk)
  * java.io
  * java.lang
  * java.math
  * java.net
  * java.time
  * java.util
* Indexing works fine IMHO without any lag
* Updated Prism to latest and included Java language highlighting
* Unfortunately, Prism doesn't seem to highlight Java code
* Added Java CSS rules, but perhaps needs extra fine-tuning and consistency with the rest


Thibaut:
Unfortunately Oracle explicitly prohibits redistribution of the Java SE documentation:  

http://www.oracle.com/technetwork/java/redist-137594.html
> Documentation - The Java SE API Specification is not redistributable.

Therefore I cannot merge this, sorry.

The only way to get Java in would be to scrape the documentation off an OpenJDK distribution, which should be covered by the GPL license.

japborst:
Sh*t, you're right... I saw them listed on the [Trello issue](https://trello.com/c/o5uswXmX/208-java-openjdk) board and didn't think it through. Might look into OpenJDK docs :)

zhongqf:
why Dash can?

Thibaut:
@zhongqf either they have an agreement with Oracle to redistribute their docs, or they're violating Oracle's copyright.

zhongqf:
@japborst I tried your fork to enable java support. but It can not download the java document with this error:
`(89/194) Java FAILED (OpenURI::HTTPError: 404 Not Found)`

It seems that it try to download "http://dl.devdocs.io/java.tar.gz" file which obviously not exist.

japborst:
@zhongqf sorry for a later reply, as I couldn't look into it earlier. Not sure if you still have the issue, as I see some additional commits on your fork. I just updated mine to the latest of upstream, so you can rebase yours to update too. I don't have any errors, but I see that you include loads more pages being scraped in your case.

A simple `/*.tar.gz/` addition in your `options[:skip_patterns]` should suffice. @Thibaut shouldn't such files be filtered in any case when scraping?

Thibaut:
@japborst the problem was that @zhongqf ran `thor docs:download` instead of `thor docs:generate`. The former downloads pre-generated docs from DevDocs's server.

PR #570, opened 2017-01-18
core/doc: cleanup slug when based on name.

kaos:
If I let the doc slug use the default, by downcasing the doc name, we might as well clean it up so it is routable.

As it is, if the name has a dash in it, it breaks if you try to load a page from the doc directly (it still works when browsing it by js navigation).

### DIFF
--- a/lib/docs/core/doc.rb
+++ b/lib/docs/core/doc.rb
@@ -48,7 +48,10 @@ module Docs
       end
 
       def slug
-        slug = @slug || name.try(:downcase)
+        slug = @slug || (
+          raise "Slug must be set explicitly when name (#{name}) consists of anything else than [\\w\\.%]" if /[^\w\.%]/ =~ name
+          name.try(:downcase)
+        )
         version? ? "#{slug}~#{version_slug}" : slug
       end
 
--- a/test/lib/docs/core/doc_test.rb
+++ b/test/lib/docs/core/doc_test.rb
@@ -55,6 +55,18 @@ class DocsDocTest < MiniTest::Spec
       doc.version = '42'
       assert_equal 'foo~42', doc.slug
     end
+
+    it "returns 'foobar' when #name has been set to 'FooBar'" do
+      doc.name = 'FooBar'
+      assert_equal 'foobar', doc.slug
+    end
+
+    it "raises error when #name has unsluggable characters" do
+      assert_raises do
+        doc.name = 'Foo-Bar'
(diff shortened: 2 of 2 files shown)


kaos:
I dropped the initial ~ in the regexp, as I realized it is used as separator for the version part of the slug, and thus should not be included here.

kaos:
I'll take a look at the failing test cases.

kaos:
Added test case to verify functionality.

Thibaut:
Hey - thanks for the PR.

I'd prefer we raise an error when the slug (generated from the name) contains illegal characters, rather than replace those characters with `_`. 

Docs with non-alphanumeric character in the name should always have an explicit `slug` defined. E.g. `C++`, `LÖVE`, `Ruby / Minitest` (in all those cases, the generated slug in this PR would be incorrect).

Thibaut:
@kaos ping

lib/docs/core/doc.rb
```diff
@@ -48,7 +48,10 @@ def name
       end
 
       def slug
-        slug = @slug || name.try(:downcase)
+        slug = @slug || (
+          raise "Slug must be set explicitly when name (#{name}) consists of anything else than [\\w\\.%]" if /[^\w\.%]/ =~ name
+          name.try(:downcase)
+        )
```
Thibaut: I think we can make this a bit simpler, like this:
```ruby
def slug
  slug = @slug || default_slug
  version? ? "#{slug}~#{version_slug}" : slug
end
```
then on L120 (after the `private`)
```ruby
def default_slug
  if name =~ /[^A-Za-z0-9_]/ 
    raise "slug must be set explicitly when name (#{name}) contains non-alphanumeric characters"
  end
  name.downcase
end
```

kaos:
I'm on vacation for another 10 days or so.. please feel free to make your
suggested modifications.
tis 21 feb. 2017 kl. 12:40 skrev Thibaut Courouble <notifications@github.com

kaos:
Hi, I'm on vacation for another 10 days or so, please feel free to make
your suggested modifications.

tis 21 feb. 2017 kl. 12:40 skrev Thibaut Courouble <notifications@github.com

Thibaut:
Merged in b36f3f8095e297e6196bb1e7268af1864312d248. Tweaked in c1ebb7a0b963cc547521743a35a97b595d9321bd. Thanks!

Issue #569, opened 2017-01-18
Add a pinned tab icon for Safari


rpowelll:
It'd be nice to add a better icon for when Safari users pin DevDocs, it currently looks like this:

<img width="115" alt="screen shot 2017-01-18 at 3 13 46 pm" src="https://cloud.githubusercontent.com/assets/320910/22050645/ca216152-dd90-11e6-9fcf-933b7bd349c5.png">

Adding an icon is fairly simple, we just need to provide a black and white SVG mask, and tell the browser what colour to use when the tab is highlighted. See [Apple's documentation](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html). This icon is also used on the new MacBook Pro's touch bar if the user has DevDocs saved as a bookmark.

Thibaut:
Hey - thanks for the feedback.

Unfortunately I have no graphic design skills (CSS is as far as my design skills go 😛) so we'll need to find someone to make us an icon.

rpowelll:
Looking at the home screen shortcut icons ([see here](https://github.com/Thibaut/devdocs/blob/master/public/images/apple-icon-160.png)), we could probably just go with a magnifying glass? If that's the case I can draw that up in Sketch pretty quickly myself, or use something like The Noun Project to source a Creative Commons licensed glyph.

Thibaut:
Fixed by #580.

PR #580, opened 2017-02-09
Added mask icon for Safari

rpowelll:
This addresses #569 and adds an SVG icon that can be used for pinned tabs in Safari, and the bookmarks shown on the touchbar on the new MacBook Pro. As a pinned tab, it looks something like this:

screen shot 2017-02-09 at 4 13 20 pm

Happy to take feedback on what a better icon would be, the current one is taken from the existing home screen bookmark icon used on iOS.


Thibaut:
Hey - thanks for the PR, and sorry for the delay (got a big update in the works). I'll get this out next week-end.

rpowelll:
No sweat 👍

Thibaut:
Merged in bde3a70480f999dba493caf3f13e1e5371bb9a6f. Thanks!

PR #585, opened 2017-02-17
Add an issue template for Github

Kaijiro:
Add an issue template for Github so that new issues warns the reporter that an issue for a new documentation set is not valid.


Thibaut:
Thanks! That's a good starting point. I'll merge and update this week-end.

Issue #593, opened 2017-03-04
OpenJDK compatibility


freayd:
Hey all,

I'm planning on adding the Java documentation using the OpenJDK 8. Before I dig into anything, I'd like to know if the [OpenJDK license](http://openjdk.java.net/legal/gplv2+ce.html) (GPL v2 with the Classpath Exception) is compatible with the DevDocs projects?

If so then I'll be submitting a pull request for it.

Thanks.

Thibaut:
Hey!

Yes, the GPLv2 is compatible with DevDocs / MPLv2.

Which source are you planning to use? Ideally we'd use pre-generated docs from somewhere. I'm on macOS and would rather not have to build OpenJDK / the docs locally if we can avoid it.

freayd:
That's great then.

Let me look a bit a the possibilities and I'll see if I can find some pre-generated documentation.

freayd:
Would that be an option for you to start from any Linux distribution package like [Debian](https://packages.debian.org/sid/openjdk-8-doc) or [Ubuntu](http://packages.ubuntu.com/yakkety/openjdk-8-doc)? Will have to check their respective licensing thought. Are you already using any Linux distribution for other documentations?

Thibaut:
Sorry, no. Every documentation is coming from a web/downloaded source, except the Ruby/Rails docs which are generated with a Ruby command-line tool.

Is there any way to turn these packages into HTML files without setting up an entire distro?

I also found http://linuxsoft.cern.ch/cern/slc63/i386/yum/updates/repoview/java-1.8.0-openjdk-javadoc.html and https://www.archlinux.org/packages/extra/x86_64/openjdk8-doc/.

FYI someone wrote a scraper for the SE documentation in https://github.com/Thibaut/devdocs/pull/560. We might be able to reuse parts of it for the OpenJDK scraper.

japborst:
Great to see people working on this! @Thibaut I have looked into OpenJDK, but the fact that there was no publicly accessible version of the docs online and some time constraints, meant I did not write a scraper for OpenJDK.

I also found this one btw: http://packages.ubuntu.com/xenial/openjdk-8-doc

@freayd feel free to fork / use my branch in any way. Personally I have no experience with OpenJDK, so I have no idea how much similarities with the Oracle docs are.

freayd:
As far as I understand, there is nothing else than sources provided by the OpenJDK project. So one option would be to download the sources and compile them which will generate the documentation.

[...]

PR #602, opened 2017-03-25
OpenJDK documentation

freayd:
As discussed in the issue #593, here comes the pull request regarding the OpenJDK documentation. This work is base on the previous #560, thanks @japborst.

The documentation is sourced from [openjdk-8-doc_8u121-b13-4_all.deb](https://packages.debian.org/sid/openjdk-8-doc). Only the subdirectoy `usr/share/doc/openjdk-8-jre-headless/api` was extracted.

I've tried to include all packages and it's working fine for me, I don't experience any slow down. They are 217 types and 46 084 entries in 4 733 files occupying 153 MB. I think many other packages than the ones proposed in #560 are useful, like `javax.xml`. What do you think @Thibaut?

The highlighting issue is solved. Unfortunately Prism remove all `<a>` links inside `<pre>` tags but I haven't looked much into it. I don't think it's a big problem thought.

The Java documentation contains a lot of duplicated information, typically a summary of all methods and further down detail explanations for these methods. I tried to remove all these duplicates so the resulting documentation is lighter and easier to read.

I have reused the logos from the request #560, but I think they might be copyrighted and I'm not sure they can be used in Devdocs. I don't see this logo on the [OpenJDK](http://openjdk.java.net/) and [IcedTea](http://icedtea.classpath.org/wiki/Main_Page) websites. And the logo used in the [Java Wikipedia article](https://en.wikipedia.org/wiki/Java_(programming_language)) seems to be subject to fair use.

### DIFF
--- a/assets/javascripts/vendor/prism.js
+++ b/assets/javascripts/vendor/prism.js
@@ -1,4 +1,4 @@
-/* http://prismjs.com/download.html?themes=prism&languages=markup+css+clike+javascript+c+cpp+coffeescript+ruby+elixir+go+json+kotlin+lua+nginx+perl+php+python+crystal+rust+scss+sql+typescript */
+/* http://prismjs.com/download.html?themes=prism&languages=markup+css+clike+javascript+c+cpp+coffeescript+ruby+elixir+go+java+json+kotlin+lua+nginx+perl+php+python+crystal+rust+scss+sql+typescript */
 var _self = (typeof window !== 'undefined')
 	? window   // if in browser
 	: (
@@ -1053,6 +1053,23 @@ Prism.languages.go = Prism.languages.extend('clike', {
 	}
 });
 delete Prism.languages.go['class-name'];
+
+Prism.languages.java = Prism.languages.extend('clike', {
+	'keyword': /\b(abstract|continue|for|new|switch|assert|default|goto|package|synchronized|boolean|do|if|private|this|break|double|implements|protected|throw|byte|else|import|public|throws|case|enum|instanceof|return|transient|catch|extends|int|short|try|char|final|interface|static|void|class|finally|long|strictfp|volatile|const|float|native|super|while)\b/,
+	'number': /\b0b[01]+\b|\b0x[\da-f]*\.?[\da-fp\-]+\b|\b\d*\.?\d+(?:e[+-]?\d+)?[df]?\b/i,
+	'operator': {
+		pattern: /(^|[^.])(?:\+[+=]?|-[-=]?|!=?|<<?=?|>>?>?=?|==?|&[&=]?|\|[|=]?|\*=?|\/=?|%=?|\^=?|[?:~])/m,
+		lookbehind: true
+	}
+});
+
+Prism.languages.insertBefore('java','function', {
+	'annotation': {
+		alias: 'punctuation',
+		pattern: /(^|[^.])@\w+/,
+		lookbehind: true
+	}
+});
 
 Prism.languages.json = {
 	'property': /"(?:\\.|[^\\"])*"(?=\s*:)/ig,
--- a/assets/javascripts/views/pages/simple.coffee
+++ b/assets/javascripts/views/pages/simple.coffee
@@ -42,6 +42,7 @@ app.views.MomentPage =
 app.views.MongoosePage =
 app.views.NginxPage =
 app.views.NodePage =
+app.views.OpenjdkPage =
 app.views.PerlPage =
 app.views.PhalconPage =
 app.views.PhaserPage =
--- a/assets/stylesheets/application-dark.css.scss
+++ b/assets/stylesheets/application-dark.css.scss
@@ -66,6 +66,7 @@
         'pages/nginx',
         'pages/node',
         'pages/npm',
+        'pages/openjdk',
         'pages/perl',
         'pages/phalcon',
         'pages/phaser',
--- a/assets/stylesheets/application.css.scss
+++ b/assets/stylesheets/application.css.scss
@@ -66,6 +66,7 @@
         'pages/nginx',
         'pages/node',
         'pages/npm',
+        'pages/openjdk',
         'pages/perl',
         'pages/phalcon',
         'pages/phaser',
--- a/assets/stylesheets/pages/_openjdk.scss
+++ b/assets/stylesheets/pages/_openjdk.scss
@@ -1,0 +1,22 @@
+._openjdk {
+  > ul.inheritance {
+    @extend %note, %note-blue;
+    li {
+      list-style: none;
+    }
+  }
+
+  ul.blockList, ul.blockListLast {
+    padding-left: 0;
+    li.blockList {
+      list-style: none;
+    }
+  }
+
+  h3 {
+    @extend %block-heading;
+  }
+  h4 {
+    @extend %block-label, %label-blue;
+  }
+}
--- a/lib/docs/filters/openjdk/clean_html.rb
+++ b/lib/docs/filters/openjdk/clean_html.rb
@@ -1,0 +1,88 @@
+module Docs
+  class Openjdk
+    class CleanHtmlFilter < Filter
+      def call
+        # Preserve internal fragment links
+        # Transform <a name="foo"><!-- --></a><bar>text</bar>
+        #      into <bar id="foo">text</bar>
+        css('a[name]').each do |node|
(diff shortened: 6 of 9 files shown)


japborst:
Very nice! I like it :-)

Thibaut:
Thanks for an outstanding PR @freayd (and @japborst for the original PR).

> I've tried to include all packages and it's working fine for me, I don't experience any slow down. They are 217 types and 46 084 entries in 4 733 files occupying 153 MB.

I'm not too worried about the speed of the search. Even with 100k entries it seems to be holding up pretty well (I did a lot of optimization a while back).

However, I'm worried about search results becoming noisy / less useful due to the 46k entries, and increasing the chance that users will hit the IndexedDB quota and lose their offline data (until browsers implement the [Storage Standard](https://storage.spec.whatwg.org/), all website data is unfortunately non-persistent and subject to deletion when e.g. the device is low on disk space).

Since I expect this doc will be used by a large number of users, we should optimize/polish it as much as we can. A good number to shoot for would be ~20k entries (IIRC the biggest doc we have today is Drupal 8, with 12k entries).

Is AWT still actively used? Packages like `java.rmi` and `javax.management` can probably be left out / not useful to most people. Packages like `javax.xml` and `javax.swing` could be extracted in separate doc versions (see docs like Haxe and Tensorflow), but let's first try to get the number down as much as we can.

> I have reused the logos from the request #560, but I think they might be copyrighted and I'm not sure they can be used in Devdocs. I don't see this logo on the OpenJDK and IcedTea websites. And the logo used in the Java Wikipedia article seems to be subject to fair use.

I don't think we should use the Java logo, since we're adding OpenJDK. Please remove the logo and I'll figure something out. Also, please rename the doc/scraper "OpenJDK".

Kaijiro:
@Thibaut maybe should we split java packages into differents documentation, so Java AWT could be apart from "Java Core" (and some additionnal splits that we should think all together).

What do you think about it ?

freayd:
Thanks all for your input.

I propose to split the documentation into four. Note that I counted only the processed HTML file sizes and not the index (db.json currently uses 74&nbsp;MB and index.json 6&nbsp;MB).

Group|Entries|Size|Packages
-------|-------|----|--------
Core|9 746|16 MB|java.io, java.lang, java.math, java.net, java.text, java.time, java.util
Tools|9 094|16 MB|java.beans, java.nio, java.security, javax.annotation, javax.crypto, javax.imageio, javax.lang, javax.management, javax.naming, javax.net, javax.print, javax.script, javax.security, javax.sound, javax.tools
GUI|18 145|26 MB|java.awt, javax.swing
Web|9 099|14 MB|java.applet, java.rmi, java.sql, javax.accessibility, javax.activation, javax.activity, javax.jws, javax.rmi, javax.sql, javax.transaction, javax.xml, org.ietf, org.omg, org.w3c, org.xml

We could combine Core and Tools which will be ~19k entries.
Swing actually relies on AWT, they are complementary so I put them together.
What do you think?

Thibaut:
@freayd again, great work! 

FYI `db.json` is just all the HTML files combined into one. It's used for the offline mode (rather than make one request for each file, which would take forever, the app downloads the entire doc in one file/request). For offline mode, the size of a doc is `index.json` + `db.json`.

Seeing that "Tools" include things like print, sound, etc. I think it makes sense to keep it separate. But we should make sure there isn't too strong a dependency between "Core" and "Tools", or the split is counter-effective (if 99% of people end up enabling both). Do the packages in "Core" cover enough of the language that most people wouldn't enable "Tools"? We should include in "Core" everything that most people need most of the time.

japborst:
@thibaut personally I think it's a fair split, but it remains subjective. Do you keep statistics of what people enable? If so, we could check and iterate once or twice.

freayd:
I've pushed a couple of commits that split the documentation into four as we talked earlier.

I've also added a new filter called clean_urls which allows to maintain links between documentations which would have otherwise been dropped. They are a few edge cases thought where we can land on a "Page not found" after clicking an inter-documentation link even though the document actually exists (refreshing the page works).

I agree and also think it is subjective to choose how to split the documentation. I've been searching for statistics about package usage but couldn't find any, so I made up some. I researched in Github "import java.lang" and did it for every of the 39 "super" packages. It's no bullet proof (packages not imported but accessed by the full path aren't counted, GitHub also warns that code queries are subject to a timeout and might be incomplete) but at least it gives a rough idea. Here are the results with the current documentation split:

Group | Package | GitHub code results | DevDocs entries
----- | ------- | ------------------- | ---------------
Core | java.util | 84 293 604 | 4 243
Core | java.io | 42 617 769 | 698
Core | java.text | 32 966 855 | 478
Web | org.xml | 27 997 701 | 325
Core | java.time | 25 434 903 | 1 400
Core | java.net | 19 803 745 | 706
Core | java.lang | 17 964 441 | 2 092
Core | java.math | 11 303 298 | 129
Web | org.w3c | 11 157 594 | 396
Web | java.sql | 8 600 354 | 919
Tools | java.security | 8 556 922 | 1 177
GUI | java.awt | 8 492 022 | 5 766
Tools | javax.annotation | 6 920 679 | 73
GUI | javax.swing | 6 463 111 | 12 379
Web | javax.xml | 6 122 457 | 2 139
Tools | java.nio | 5 419 356 | 1 040
Tools | java.beans | 5 214 548 | 468
Tools | javax.print | 4 718 565 | 1 078
Tools | javax.net | 4 217 379 | 315
Tools | javax.lang | 3 504 900 | 433
Tools | javax.security | 2 700 837 | 359
Tools | javax.tools | 2 312 077 | 167
Web | javax.sql | 2 025 446 | 527
Web | org.ietf | 1 433 874 | 138
Web | javax.transaction | 1 315 190 | 64
Tools | javax.management | 1 064 609 | 1 532
Web | java.rmi | 920 072 | 296
Tools | javax.script | 731 676 | 127
Web | java.applet | 724 655 | 49
Tools | javax.crypto | 701 461 | 251
Tools | javax.naming | 700 389 | 619
Web | org.omg | 616 352 | 3 597
Tools | javax.imageio | 552 398 | 876
Web | javax.activity | 447 999 | 7
Web | javax.rmi | 354 132 | 92
Tools | javax.sound | 287 089 | 579
Web | javax.activation | 273 505 | 119
Web | javax.jws | 159 454 | 64
Web | javax.accessibility | 145 285 | 367

Although it is arbitrary, we might want to integrate the most used "Tools" packages into "Core", like the security related ones (**java.security, javax.security, javax.crypto**), the counterpart of the already "Core" packages (**java.nio, javax.net, javax.lang, javax.tools**) and maybe the other couple of widely used packages (**javax.annotation, java.beans**). In this case, "Core" will have 14&nbsp;029 entries and "Tools" 4&nbsp;811.

Thibaut:
> Do you keep statistics of what people enable? If so, we could check and iterate once or twice.

I can extrapolate how popular a doc is based on page views tracked client-side, but I don't have precise stats on what people enable.

---

Great research @freayd. Let's remove the "Tools" package for now (put it all in "Core"). I'll try to trim the markup and reduce the no. of entries a bit more when I merge this, + do some performance testing.

I'm catching up with doc updates today, then traveling for a week, so won't be able to get this out before a couple weeks from now.

Please squash your commits, so I can cherry-pick locally (fits my workflow better than merging on github.com).

freayd:
Sorry for the delay, it should be ready to merge now.

freayd:
Sorry for the delay. I've merged Core & Tools and fixed a regression from the last commit. Should be ready to merge now.

Thibaut:
Started working on this yesterday but didn't have time to finish. Will try to finish next week-end.

Thibaut:
Done: f606c0abc8fce8c3e0379becfc264c01f1a726f6 2efce745210f354d4226cf78822791e5f9abad89 🎉 
http://devdocs.io/openjdk~8/
http://devdocs.io/openjdk~8_gui/
http://devdocs.io/openjdk~8_web/

I managed to reduce the size of the docs by ~30% by trimming the markup as much as possible.

I also removed "constructor" entries, since they always have the same name as the page's name, and "field" and "constant" entries, because from what I can tell most of them aren't very useful and they're located at the top of the page, so should be easy to find. Overall this reduces the number of entries by ~20%. I might re-add the "field" and "constant" entries if people ask for it and once I've had time to optimize the search some more (didn't have time to do it for this PR).

Thanks for the hard work!

freayd:
Very nice job, many thanks ! 🎉

PR #605, opened 2017-03-29
Responsive browser compatibility tables

lex111:
[Source page](http://devdocs.io/javascript/global_objects/int8array)

The table does not fit completely into the screen on mobile, and a horizontal scroll appears:
image

The solution is to add `overflow-y: auto;` (look more [here](http://maxdesign.com.au/articles/simple-responsive-table/)) to the container with the table, then the horizontal scroll will be only at the container, and not at the page:
image

_Note: above in the screenshot you can not see the horizontal bars, but they are when you hover._

Probably hurried with the PR. Can be so conceived, but it looks like a bug.

### DIFF
--- a/assets/stylesheets/components/_mobile.scss
+++ b/assets/stylesheets/components/_mobile.scss
@@ -105,6 +105,12 @@
     max-width: none;
     margin-left: 0;
   }
+  
+  // Responsive browser compatibility tables
+  #compat-desktop,
+  #compat-mobile {
+    overflow-x: auto;
+  }
 }
 
 //


japborst:
Interestingly your text talks about adding `overflow-y: auto` but your commit adds `overflow-x: auto`. 

Why only the comparability tables and not all tables? You make use of auto scrollbars, after all.

lex111:
@japborst In the description, I made a mistake, in the commit, the correct change, `overflow-x`, of course.

For all tables it does not work, because It is necessary that the table be wrapped in a separate container (div tag), to which this overflow-x will be applied, then the table inside will have the ability to scroll.

Thibaut:
Merged in 302e3f1afecc9df4043d501dc9c17b1c3755adb1. Thanks!

PR #303, opened 2015-11-27
Added Dockerfile

cahna:
```
docker build -t devdocs/devdocs .
docker run -d -p 9292:9292 devdocs/devdocs
```

PR #623, opened 2017-05-08
Improve Dockerfile, and added alpine version

alexbrazier:
- Added docker ignore to remove unnecessary files
- Reduce number of layers, and optimise build speed (e.g. don't re-download gem files on every file change)
- Add alpine version to reduce size
- By deleting `/tmp` it reduces the size from ~4GB to ~2GB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,13 @@ WORKDIR /devdocs
 
 RUN apt-get update && \
     apt-get -y install git nodejs && \
-    gem install bundler
+    gem install bundler && \
+    rm -rf /var/lib/apt/lists/*
 
 COPY Gemfile Gemfile.lock Rakefile /devdocs/
 
-RUN bundle install --system
+RUN bundle install --system && \
+    rm -rf ~/.gem /root/.bundle/cache /usr/local/bundle/cache
 
 COPY . /devdocs
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,21 @@
-
 FROM ruby:2.4.1
 MAINTAINER Conor Heine <conor.heine@gmail.com>
 
-RUN apt-get update
-RUN apt-get -y install git nodejs
-COPY . /devdocs
-RUN gem install bundler
-
 WORKDIR /devdocs
 
+RUN apt-get update && \
+    apt-get -y install git nodejs && \
+    gem install bundler
+
+COPY Gemfile Gemfile.lock Rakefile /devdocs/
+
 RUN bundle install --system
-RUN thor docs:download --all
+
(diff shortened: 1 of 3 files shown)


alexbrazier:
@Thibaut Any chance you could also setup an automated build on https://hub.docker.com so that they are automatically updated when the code is changed?

Dockerfile
```diff
@@ -1,17 +1,21 @@
...
-RUN apt-get update
-RUN apt-get -y install git nodejs
-COPY . /devdocs
-RUN gem install bundler
-
 WORKDIR /devdocs
 
+RUN apt-get update && \
+    apt-get -y install git nodejs && \
+    gem install bundler
```
amywoodehy: what about? `&& rm -rf /var/lib/apt/lists/*`

alexbrazier: @amywoodehy I have added that change, along with a couple of others

Dockerfile
```diff
@@ -1,17 +1,23 @@
-
 FROM ruby:2.4.1
 MAINTAINER Conor Heine <conor.heine@gmail.com>
```
Thibaut: What is the `MAINTAINER` field used for? (not familiar with Docker)
The `Dockerfile` was originally added in https://github.com/Thibaut/devdocs/pull/303. At this point we should probably change it to my name/email, or remove it.

Thibaut:
> Any chance you could also setup an automated build on https://hub.docker.com so that they are automatically updated when the code is changed?

I'll set that up next week-end.

Dockerfile
scherepn: MAINTAINER is just additional metadata used to identify the owner of the image. It doesn't affect behavior, nor is it used by Docker when the container is run.

Use of MAINTAINER is sort of deprecated though. It's recommended practice to use LABEL with MAINTAINER. i.e:

    LABEL MAINTAINER="Connor Heine <connor.heine@gmail.com>"

It doesn't really matter in this specific case because the images don't have any other labels, but the above method is preferred because multiple labels can be added to an image in a single layer whereas the dedicated MAINTAINER will cause two layers to be created.

alexbrazier: I've removed the maintainer layer from both dockerfiles

alexbrazier:
@Thibaut Awesome

I've attached a screenshot of the config to build both the alpine image and the regular:
image

To setup the repo you just need to select create an automated build, link github and select the repo, then enter the config above
image

Thibaut:
Done: 50ca8e26455d38289fffe39f4c441d61ac07b411
https://hub.docker.com/r/devdocs/devdocs/

Thanks for the PR & instructions!

PR #684, opened 2017-10-16
Add bluebird docs

brewingcode:
It's only now that I'm sending the PR that I realize there was a [previous PR](https://github.com/Thibaut/devdocs/pull/286) that I really could have used a starting point. :-) Oh well, I need to practice my ruby anyway. The `get_type` method is a complete hack, because the Bluebird docs do not include *ANY* categorization hints at the individual doc pages: not in the url, and not in the markup.

image


lib/docs/filters/bluebird/entries.rb
```diff
@@ -0,0 +1,66 @@
...
+          else
+            slug_end = slug.sub(%r(^docs/api/), '')
+            if v.include?(slug_end.downcase)
+              type = k
+              break
+            end
+          end
+        end
+
+        type.to_s
```
brewingcode: For a non-rubyist, what exactly is going on with the keys of the `TYPE_MAP` hash? Both here and [up above on line 49](https://github.com/Thibaut/devdocs/pull/684/files#diff-5b23a8771aed8bfa224b7f36bca446f7R49) the `to_s` was required to get things working the way I would naively expect:

* `if k.casecmp(name.strip)` would NEVER equal 0, even when `puts` was telling me that `name.strip` was a key that was _right there_ in the `TYPE_MAP` hash, like "Core"

* returning just `type` was showing up in the `thor docs:generate` as a single type instead of the 20-ish I would expect (one type per key in `TYPE_MAP`)

Thibaut:
Thanks for the PR! I'll look into this as soon as I can (could be a few weeks). If you don't mind just squash your commits please.

brewingcode:
Commits squashed into 24add008b

clakech:
Nice PR. I want this one to be merged

I tried your branch but I got an issue running `thor docs:generate bluebird --force --debug`
```
Queue:   bluebirdjs.com/docs/api-reference.html
Process: bluebirdjs.com/docs/api-reference.html                                                                                                                     [14ms]
ERROR:
  http://bluebirdjs.com/docs/api-reference.html
  Docs::ContainerFilter::ContainerNotFound: element 'body .post' could not be found in the document, url=http://bluebirdjs.com/docs/api-reference.html

**Comment by brewingcode** [comment]
> I tried your branch but I got an issue running thor docs:generate bluebird --force --debug
> ` Docs::ContainerFilter::ContainerNotFound: element 'body .post' could not be found in the document`

@clakech I kept seeing this intermittently as well, but it always went away eventually (as you saw, I think). I chalked it up to the bluebirdjs.com server responding with an occasional error due to the rapid scraping or something, though I never did confirm.

> when I click on a link in the main area, a new tab is opened in the browser

Hmmm, yeah I see this too, and it only seems to happen when the right-side window is showing the main bluebird TOC _when it is disabled_. If you click "enable" in the left-side, then swap the TOC out for a specific page, and then go back to the TOC, the links stay in the tab.

And further, once Bluebird is enabled, there is no issue with the links opening new tabs.

There's something strange going on there that I don't completely understand, maybe @Thibaut has an idea...?

File: lib/docs/filters/bluebird/entries.rb (hunk 1 above)
Comment (j-f1) [reply lib/docs/filters/bluebird/entries.rb]: The keys of the `TYPE_MAP` hash are [`Symbol`](http://devdocs.io/ruby~2.4/symbol)s, which are similar to strings but not the same. If you want the hash to have string keys, something like this should work:

```ruby
TYPE_MAP = {
  'Core' => %w(new-promise then spread ...),
  'Synchronous inspection' => %w(promiseinspection isfulfilled isrejected ...)
}
```

brewingcode: Oh, the operator _after_ the key determines whether the key is a symbol or string. Thanks @j-f1: I was able to figure out the fact that the keys were ending up symbols instead of strings, but could not figure out why.

Is it more accepted to separate keys and values in a hash declaration fat arrows rather than colons in Ruby? I see code elsewhere doing `{ :key => "value" }`, which makes it immediately clear what is going on. I can also imagine that `{ key: "value" }` is a nicer way of writing the exact same thing, though.

j-f1: > Is it more accepted to separate keys and values in a hash declaration fat arrows rather than colons in Ruby?

It’s the best option if you want the keys to be symbols. If you want them to be strings, you have to use the hash rocket (`=>`) syntax.

brewingcode: > hash rocket 

*There* is the term I want to google. Thanks again, @j-f1.

Thibaut:
> > I tried your branch but I got an issue running thor docs:generate bluebird --force --debug
> > ` Docs::ContainerFilter::ContainerNotFound: element 'body .post' could not be found in the document`
>
> @clakech I kept seeing this intermittently as well, but it always went away eventually (as you saw, I think). I chalked it up to the bluebirdjs.com server responding with an occasional ~error~ malformed response due to the rapid scraping or something, though I never did confirm.

I've seen this happen when the scrapped website can't handle the load (to work around that, you can lower the concurrency of requests [here](https://github.com/Thibaut/devdocs/blob/master/lib/docs/core/requester.rb#L22)), or due to memory consumption issues by the image processing code (which unfortunately I haven't been able to figure out :/).

> > when I click on a link in the main area, a new tab is opened in the browser
>
> Hmmm, yeah I see this too, and it only seems to happen when the right-side window is showing the main bluebird TOC _when it is disabled_. If you click "enable" in the left-side, then swap the TOC out for a specific page, and then go back to the TOC, the links stay in the tab.
>
> And further, once Bluebird is enabled, there is no issue with the links opening new tabs.
>
> There's something strange going on there that I don't completely understand, maybe @Thibaut has an idea...?

This behavior is intended. When a doc is disabled, the app is unable to display any page of that doc except the root page (since it doesn't have the metadata with all the url/page mappings).

clakech:
@brewingcode any news on this one ?

brewingcode:
I'm stuck on the intermittent `Docs::ContainerFilter::ContainerNotFound` error, even after reducing `:max_concurrency` to 1. From my limited debugging, it appears the `doc` that typhoeus is returning is completely bogus: if I `puts doc.to_html` at the `raise` here:

https://github.com/Thibaut/devdocs/blob/master/lib/docs/filters/core/container.rb#L12

...it's nothing but a single non-utf8 character.

Thibaut:
Done: b399a0ca6f3df771888247077ef517867de91469 dac419289bea756e6b697c82572e5ebe028cceb3
https://devdocs.io/bluebird/ 

> I'm stuck on the intermittent `Docs::ContainerFilter::ContainerNotFound` error, even after reducing `:max_concurrency` to 1.

This can also happen when the server incorrectly returns GZIP content for non-GZIP requests. Adding `self.force_gzip = true` (which forces the scraper to always use GZIP) seems to have done the trick.

brewingcode:
Thanks @Thibaut!

thgreasi:
Woohoo!

PR #686, opened 2017-10-19
Minor cosmetics for PHP

KalleZ:
This change contains two, relatively small changes:

1) The release of the PHP documentation is always the latest version of PHP, and even in some cases future releases such as the upcoming 7.2
2) The source location for PHP is actually our own git host at git.php.net, whereas the Github repository acts as a mirror

### DIFF
--- a/lib/docs/scrapers/php.rb
+++ b/lib/docs/scrapers/php.rb
@@ -4,7 +4,7 @@ module Docs
 
     self.name = 'PHP'
     self.type = 'php'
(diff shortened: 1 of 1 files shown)


Thibaut:
Thanks for the PR. Regarding 1), I'd prefer keeping it as is, as this gives an indication of when the documentation was last updated. Maybe I'll just remove the "up to" part.

If you'd like to get this merged in, please update your commit and I'll get it in next time I update the PHP doc.

PR #704, opened 2017-10-24
add brew doc (1.3.6)

clakech:
I still have an issue with the https://docs.brew.sh/Installation.html that is downloaded in 1ms only and seems buggy... Don't know why ? Cache ? This page works fine on a browser


Thibaut:
Done: 438fa8e6d04486fde5901ddb916ea09074d38e78 b8c8a42f868fa261b26d24bd25fd48af29139a33
https://devdocs.io/homebrew/

PR #705, opened 2017-10-25
Adds Scala documentation

jamiely:
Specifically, api docs for Scala 2.10, 2.11, and 2.12.


Thibaut:
Thanks for the PR! I have a bit of a PR backlog (and need to take care of updates as well), so it might be a few weeks before I'm able to work on this.

I haven't tried this yet, but looking at the code, we need to find a way to clean up the markup even more, such that the documentation needs almost no custom CSS. Doc-specific CSS is hard to maintain and reduces the consistency between docs, which is a big part of what makes the reading experience on DevDocs good.

Removing all unnecessary markup also helps with performance (not just when loading pages, but also for offline mode, etc.). And currently your PR only works with the light/default theme. Since you hardcoded colors in the CSS, it doesn't work when the dark theme is enabled (another reason why we don't want doc-specific CSS).

If you don't mind, try to update your code to minimize the amount of markup in the output (e.g. remove all unnecessary `<div>` / elements that you have set to `display: none`) and reduce the need for custom CSS.

jamiely:
Oh okay, will do. Thanks for the feedback.

jamiely:
I've updated the PR per your comments.

tasgoon:
Any update on this?

jmerle:
Hey! Thanks for your contribution and sorry for the insane delay on this, but better late than never I guess.

I removed versions <2.12 and added the docs for Scala 2.13. Since the docs prior to 2.11 are styled differently, this makes it way easier to maintain the scraper, and since Scala 2.12 was released in 2016 I don't think 2.11 is still needed. I also refactored parts of the filters to generate docs more consistent with the rest of the docs available on DevDocs.

The docs for Scala 2.12 and 2.13 are now available on [devdocs.io/scala~2.12_library](https://devdocs.io/scala~2.12_library/) and [devdocs.io/scala~2.13_library](https://devdocs.io/scala~2.13_library/) respectively (replace `library` with `reflection` for the reflection docs).

PR #739, opened 2017-12-26
Update Fish documentation (2.7.1)

maxnordlund:
It was relatively recently released https://fishshell.com/release_notes.html

### DIFF
--- a/lib/docs/scrapers/fish.rb
+++ b/lib/docs/scrapers/fish.rb
@@ -17,6 +17,11 @@ module Docs
       Licensed under the GNU General Public License, version 2.
     HTML
 
+    version '2.7' do
+      self.release = '2.7.1'
+      self.base_url = "https://fishshell.com/docs/#{version}/"
+    end
+
     version '2.6' do
       self.release = '2.6.0'
       self.base_url = "https://fishshell.com/docs/#{version}/"


Thibaut:
Done in 6700c5e30c9f4a6c026dd63395bba1494736b181.

maxnordlund:
I'm curious as to why you didn't just merge the PR?

Thibaut:
Because this kind of PR is not useful, and the contributing guidelines specify not to submit those: https://github.com/Thibaut/devdocs/blob/master/CONTRIBUTING.md#updating-existing-documentations

You didn't say if the scraper still works after the change, so this saves me no time. It's faster for me to do the update through my usual workflow. In the future you can request updates [here](https://trello.com/c/2B0hmW7M/52-request-updates-here) (or send PRs when those updates aren't straightforward).

maxnordlund:
I see, well then i does make a bit more sense. You right, I din't verify the scraper, but as the website hadn't changed it's structure I figured it would work. Actually the docs for fish hasn't changed format in the years I've used it, which I find helpful for many reasons, scrapers included.

Next time a new contributor comes along and makes a (relatively) small misstake like this I gently suggest linking to the guidelines. It would have help me understand straight away instead of needing to ask.

Since this seems like something that a fair amount of people may trip on, do you think it should also be mentioned in the README? In that case I can make a PR updating it.

Thibaut:
@maxnordlund You're right. Sorry I didn't provide an explanation.

> do you think it should also be mentioned in the README?

I'd rather keep all contributing guidelines in the CONTRIBUTING file, since it's linked in various places in the GitHub UI.

maxnordlund:
True, but this is akin to a FAQ, so a bit of repetition may be prudent. Anyway, sounds good, and thanks for the apology I appreciate it.

PR #740, opened 2017-12-29
Update OpenSearch instructions for Firefox

metarmask:
Relevant article: ["How to search IMDB, Wikipedia and more from the address bar"](https://support.mozilla.org/en-US/kb/how-search-from-address-bar)


Thibaut:
Merged in 6baa21ee0bb759956ffb8612602fef86645d1a03. Thanks!

PR #773, opened 2018-03-25
Add WordPress documentation

LeGaS:
This PR adds docs for [WordPress](https://wordpress.org/), which [has been requested](https://trello.com/c/mDIVCppx/85-wordpress) ≈582 times.  The library is [licensed](https://github.com/WordPress/WordPress/blob/master/license.txt) under the GPL2 license.

The documentation is based on version 4.9.4 and uses the WordPress [Code Reference](https://developer.wordpress.org/reference/). The documentation is generated from Functions, Hooks and Classes, I’ve omitted the Methods topic, because the links are scraped through the individual classes in the Classes topic.

This documentation does not contain any of the [WordPress Codex](https://codex.wordpress.org) pages.

I don’t work with Ruby and only have little experience with it, so please forgive any mistakes I made.

There were two things I could not solve:
* Every page is generated to a folder instead of an HTML file with the name of the function / class / etc. Although it works, it bothers me. I’ve tried it with the `trailing_slash` option turned off, but it generated duplicate pages. 
* Fixed, this is working now.
~Syntax highlighting is only working with the “Source” section of a page, but not for the “More information”, e.g.: [add_image_size()](https://developer.wordpress.org/reference/functions/add_image_size/) - in the “Source” section, the code is wrapped in a `.source-code-container` div, but in “More information” it is not, I guess that’s why it is not working.~

Looking forward to your feedback.


Thibaut:
Hey there - sorry for the lack of response. I'm currently not able to devote much time to DevDocs beyond doing some documentation updates. We're working with freeCodeCamp to try to find new active maintainers — hopefully soon :)

jukefr:
Any news on this ?

jmerle:
I don't have news on this particular PR, but I do know that Thibaut is in the process of handing everything over to freeCodeCamp and making it easier to contribute, which was (is?) planned to be done by the end of 2018.

After that's done, @beaucarnes is going to be the man in charge. Although I don't know about any plans yet, I'm guessing PR's will be reviewed old-to-new, which means this would be reviewed in January or February.

lib/docs/scrapers/wordpress.rb
```diff
@@ -0,0 +1,39 @@
+module Docs
+  class Wordpress < UrlScraper
+    self.name = 'WordPress'
+    self.type = 'wordpress'
+    self.release = '4.9.4'
+    self.base_url = 'https://developer.wordpress.org/reference/'
```
jukefr: The trailing `/` should be removed.
It results in : 
```
© 2003–2018 WordPress Foundation
Licensed under the GNU GPLv2+ License.
https://developer.wordpress.org/reference//
```

lib/docs/scrapers/wordpress.rb
```diff
@@ -0,0 +1,39 @@
+module Docs
+  class Wordpress < UrlScraper
+    self.name = 'WordPress'
+    self.type = 'wordpress'
+    self.release = '4.9.4'
```
jukefr: maybe update to latest ?(4.9.8 or even 5.0.0 by the time this gets merged)

lib/docs/scrapers/wordpress.rb
```diff
@@ -0,0 +1,39 @@
...
+    ]
+
+    options[:skip_patterns] = [
+      /\Afunctions\/page\/\d+/,
+      /\Ahooks\/page\/\d+/,
+      /\Aclasses\/page\/\d+/
+    ]
+
+    options[:attribution] = <<-HTML
+      &copy; 2003&ndash;2018 WordPress Foundation<br>
```
jukefr: Update for 2019 ?

mitchellknight:
What needs to be done for this to deployed and closed?

j-f1:
We’re still working on the process Thibaut mentioned. Also looks like @jukefr left some suggested changes above.

cycle337:
what's hindering progress here ? I'm offering to help

jmerle:
Sorry for the delay and thanks for contributing! I've made a few changes to the scraper and it's filters to make them easier to maintain and to ensure the generated docs look like the other available docs (there were some header problems due to the h1 being nested inside other elements and the header not always coming first). Also updated the version, logo and footer to their latest versions.

The WordPress docs are now live on [https://devdocs.io/wordpress/](https://devdocs.io/wordpress/).

PR #778, opened 2018-03-26
Update haxe.rb

markknol:
Adds all targets / copyright year.

Thanks for this awesome tool, keep up the good work!!


Thibaut:
Merged in 2c5f9d0627d3707d8f4d46239effe6ccb211930a. Haxe removed the Macro docs but all the others are live on devdocs.io.

markknol:
Thanks!

I'm mistaken because macro is a Haxe target but the sources are in the haxe package, so its all good!

PR #810, opened 2018-05-24
Add pygame documentation

### LINKED ISSUE #807: Pygame
I'm going to have a go at adding a scrapper for `pygame`. I've never used ruby before, though...

math2001:
Add the pygame documentation.

There are a few things that I'm not too sure about...

### HTML semantic

The pygame documentation is sphinx link, but messier, so the default CSS doesn't highlight `dt` properly (that's because there are `dl` nested in `dl`). So, I manually replace each `dt` with `h3` and everything looks nice and neat (with `self.type = 'simple'`). The advantage is that there's no new CSS, but semantically, the HTML is **ugly**. But for devdocs, I don't think it matters too much, no?

### The entries

I remove the `pygame.` prefix that there is on every entry. The problem is that the function `pygame.init()` for example is listed as `init()` in the index... So, there's 3 possibilities:

- leave as is: no `pygame.` at all
- put `pygame.` everywhere
- put `pygame.` only for the module `pygame` itself. That brings inconsistency though, but might be a bit clearer.

I think that's it though...

Thanks for devdocs.io anyway :+1:

---

#### FileScrap

The pygame documentation limits the number of request to 200 a day, so if you want to tweak around, it might be limiting (it was for me at least), so I've made a bash script that scraps the pages i need, and then I can use a `FileScraper` instead. Just in case you need to tweak some things, here it is:

<details>
<summary>
Just run this script
</summary>

```bash
mkdir ref

urls="py-modindex.html
ref/color.html
ref/cursors.html
ref/display.html
ref/draw.html
ref/event.html
ref/font.html
ref/image.html
ref/joystick.html
ref/key.html
ref/locals.html
ref/mask.html
ref/mixer.html
ref/mouse.html
ref/rect.html
ref/surface.html
ref/sprite.html
ref/time.html
ref/transform.html
ref/bufferproxy.html
ref/freetype.html
ref/gfxdraw.html
ref/midi.html
ref/overlay.html
ref/pixelarray.html
ref/pixelcopy.html
ref/sndarray.html
ref/surfarray.html
ref/camera.html
ref/cdrom.html
ref/examples.html
ref/math.html
ref/music.html
ref/pygame.html
ref/scrap.html
ref/tests.html"

curl `printf 'https://www.pygame.org/docs/%s -o %s ' $(echo -e "$urls" | awk "1;1")`
```

Fixes #807.


beaucarnes:
This resolves #807

Thibaut:
Merged in 6f4f3485a4939edfb3c7c658b3510f49806ac741 (removed changes to `public/docs/docs.json`). Finished in 34fe833d82c11f4e64e251d7e0992756318a00ec.

https://devdocs.io/pygame/

Thanks for the great PR!

math2001:
Thanks 👍

PR #814, opened 2018-05-28
Upgrade Ansible doc to version 2.5

tinproject:
Upgrade to support Ansible 2.5 new documentation structure.

Notes: The new doc structure does not highlight the current module type on the toc, so that 'module category' present on previous versions is not available (or at least I do not know how to obtain it). All the modules goes all under the same `Modules: Index`.

### DIFF
--- a/lib/docs/filters/ansible/entries.rb
+++ b/lib/docs/filters/ansible/entries.rb
@@ -11,14 +11,28 @@ module Docs
       end
 
       def get_type
-        if slug.include?('module')
-          if name =~ /\A[a-z]/ && node = css('.toctree-l2.current').last
-            "Modules: #{node.content.remove(' Modules')}"
+        if slug =~ /\Acli\//
+          'CLI Reference'
+        elsif slug =~ /\Anetwork\//
+          'Network'
+        elsif slug =~ /\Aplugins\//
+          if name =~ /\A[a-z]/ && node = css('.toctree-l3.current').last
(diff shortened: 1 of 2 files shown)


beaucarnes:
I tested this locally and it works great. @Thibaut On the Trello board there was a request to keep version 2.4. This does not keep 2.4. Do you think both should be available or just 2.5?

tinproject:
I do not keep the 2.4 version because I thought that with the structure change will need different scrapers to maintain.

Also, I don't believe keeping the old version adds any value, the original module docs already contains notes on at which version the module or a field was added ([an example here](http://docs.ansible.com/ansible/latest/modules/openssl_certificate_module.html#openssl-certificate-module)).

My main usage of the devdocs' Ansible doc. is to quickly access and read module docs. For other uses I don't mind to take a look at the Ansible documentation page directly.

Thibaut:
Thanks! I'll take a look this week-end. We should indeed keep the 2.4 version around. One of the main benefits of DevDocs is that it lets you access multiple versions of a documentation, including older ones, rather than just the latest one.

beaucarnes:
@tinproject I just created a PR on your PR to add the 2.4 version in addition to the 2.5 version. I think if you accept my PR, it will show up here as well.

Thibaut:
Merged in bb36df7085ee74a744052a4577737edf9dab9d13. Thanks!

Added back the 2.4 doc in bcb544545290e381104b1241db9e2caf4691e19f.

PR #820, opened 2018-06-10
Use browser dimensions for mobile detection

### LINKED ISSUE #634: It seems that the displayed layout is for mobile when I am on my laptop
Hi,

Since a couple of weeks, I noticed a change in the layout. But I think it is actually unintended. This new layout is quiet unusable with the keyboard... I actually think that my setup is recognized as a mobile, hence displaying a different layout. What I see in Firefox:

capture d ecran de 2017-06-06 10-52-33

What I see in Chromium:

capture d ecran de 2017-06-06 10-52-41

Browser: Firefox 53.0.3
OS: Linux Ubuntu Gnome 16.04.2

grantbdev:
This PR resolves #634.

It seems that Firefox in some situations incorrectly reports the device dimensions. The users who reported the issue with Chromium on Linux might be having a problem with their window managers incorrectly reporting the display dimensions as one user suggested.

Outside of those weird bug areas, this PR also makes DevDocs a little [more responsive to browser resizing](https://developers.google.com/web/fundamentals/design-and-ux/responsive/#a_note_on_min-device-width), but since the mobile detection is done on load the user would need to refresh the page for the layout to transition after browser resizing.

### DIFF
--- a/assets/javascripts/views/layout/mobile.coffee
+++ b/assets/javascripts/views/layout/mobile.coffee
@@ -16,8 +16,8 @@ class app.views.Mobile extends app.View
   @detect: ->
     try
       (window.matchMedia('(max-width: 480px)').matches) or
-      (window.matchMedia('(max-device-width: 767px)').matches) or
-      (window.matchMedia('(max-device-height: 767px) and (max-device-width: 1024px)').matches) or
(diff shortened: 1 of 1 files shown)


Thibaut:
Merged in d5d913fc11460bdc3fc3483f1a45921b7cc6d6b7. Thanks!

PR #821, opened 2018-06-12
Use `document.scrollingElement` for mobile scrolling

### LINKED ISSUE #769: Anchors are broken in mobile mode
I use devdocs.io in a tiny Firefox window so it ends up in mobile mode. Some parts of documentation have faulty shortcuts in this mode.

To reproduce:
1. Firefox on Android or small Firefox window
2. Enable docs for Crystal
3. Find a method you want to learn about, like Enumerable's reduce. Click on the link.
4. Get sent to the top of the Enumerable page instead of the specific method.
5. That's fine, scroll down to the table of contents and click #reduce there.
6. Nothing happens.

Expected result:
Get sent to the part of the page that explains the specific method. Also, for the table of contents links to do the same thing. These two things may be related.

This works fine in desktop mode, not sure why.

grantbdev:
This PR resolves #769.

As the commit message explains, the underlying issue is [browser compatibility and shifting web standards](https://miketaylr.com/posts/2014/11/document-body-scrollTop.html). [Document.scrollingElement](https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement) fills in the gap for modern browsers, but the support for that is [not there for some of the minimum versions](https://developer.mozilla.org/en-US/docs/Web/API/document/scrollingElement) listed in this project's README so I've kept the pre-existing value as a fallback when `document.scrollingElement` is undefined.

### DIFF
--- a/assets/javascripts/views/content/content.coffee
+++ b/assets/javascripts/views/content/content.coffee
@@ -19,7 +19,10 @@ class app.views.Content extends app.View
     after:  'afterRoute'
 
   init: ->
-    @scrollEl = if app.isMobile() then document.body else @el
+    @scrollEl = if app.isMobile()
+      (document.scrollingElement || document.body)
+    else
+      @el
     @scrollMap = {}
     @scrollStack = []
 


Thibaut:
Merged in d657377d1c52c6140b24b3ac278d7670863d230a. Thanks!

Issue #791, opened 2018-04-13
Don't lowercase case-sensitive attributes


smilingthax:
The SVG documentations search result list shows all attributes as lowercase. 
But to work properly, they have to be correctly capitalized. 
In my case, the search result showed `viewbox`, whereas the doc page itself used `viewBox`. Only the latter one was recognized by the svg renderer.

Please don't mangle case-sensitive names in the search result list.

Thibaut:
#822.

PR #822, opened 2018-06-12
Fix SVG case issue (#791)

beaucarnes:
SVG attributes now keep correct case. 

I wonder why the attributes were changed to lowercase to begin with. Maybe there is a good reason I don't know about.


Thibaut:
Merged in 55a77f633fbd08784b6b34696d716015464d3a60. Thanks!

PR #830, opened 2018-07-03
Add Bash documentation

jmerle:
Added a Bash scraper, scraping the [Bash Reference Manual](https://www.gnu.org/software/bash/manual/html_node/index.html) as suggested in the [Trello card](https://trello.com/c/xd4ZQe6P).

The documentation is licensed under the [GNU Free Documentation License](https://www.gnu.org/licenses/fdl-1.3.en.html). I've never heard about it before, but it seems to permit alteration, redistribution and commercial use.

The logo comes from [odb/official-bash-logo](https://github.com/odb/official-bash-logo) and is licensed under the [Free Art License](http://artlibre.org/licence/lal/en/).

Types:

And thanks for creating DevDocs, it was my main motivation to learn Ruby (great documentation and the chance to contribute to something I use daily).


Thibaut:
Merged in 1d56144a4b71a0c1de422914e4ed2c66f14194c8. Finished in a71f48ea10a012f3ac8103991bfb86eeae68e117.

Amazing PR overall—thanks!

Here's a quick rundown of the changes I made:

- Undid the changes `.gitignore`.
- Squashed commits.
- Removed the need for custom CSS for the lined headings by making the first node always an `<h1>` in the `CleanHtmlFilter`.
- Simplified CSS (+ changed to use CSS color variable so that it works in dark mode).
- Removed some `class` attributes to reduce file size.
- Changed `<tt>` to `<code>` so that the CSS applies.
- Removed nested nodes inside `<pre>` (pretty much always have to do that)

https://devdocs.io/bash/

PR #831, opened 2018-07-04
Hide sidebar when mouse is outside the window

### LINKED ISSUE #829: Do not show sidebar when mouse is outside the window
Assume a typical development environment: two monitors, code editor or IDE on the left, web browser and various chat applications on the right. Assume also that the “Automatically hide and show the sidebar” option is set. In this scenario, the behavior of the sidebar is distracting and annoying.

Steps to reproduce:

1. Set up the window configuration as described above. In a single monitor case, tile the editor and the browser side-by-side.
2. In Preferences, check the “Automatically hide and show the sidebar” option.
3. Search for some term.
4. Click in the code editor to focus it.

Desired behavior: Can work with the code, glancing at the documentation from time to time.

Observed behavior: When the mouse pointer is within the code editor window and thus to the left of the web browser displaying Devdocs, the sidebar pops up, hiding part of the documentation.

I suggest that the sidebar should only pop up when the mouse is actually within the left margin of the page, and hide when the mouse goes outside the window.

jmerle:
Fixes #829.

Before (yes, "Automatically hide and show the sidebar" is toggled on here):

After:


yurikhan:
Can this be rolled out please?

PR #846, opened 2018-07-24
Add Qt documentation

Nicolapps:
This pull requests adds a scraper for the Qt documentation ([Trello card](https://trello.com/c/Yamxt4EM)). The latest version (5.11) is supported, as well as the 5.9 and 5.6 LTS versions.

The documentation is licensed under the [GNU Free Documentation License v1.3](https://www.gnu.org/licenses/fdl.html). The utilisation of the logo [is allowed](http://brand.qt.io/guidelines/) as it’s only used for informational purposes.

The scraper uses a UrlScraper because it’s difficult to build locally the documentation to HTML. It generates a huge documentation set (~3K pages and ~36K entries), but it only reflects the size of the Qt framework. The search engine is still working perfectly and splitting the documentation into multiple versions would only make it more difficult to use.

Types: (1 type by Qt module)
Qt documentation types


j-f1:
How do you feel about removing “Qt” from before each topic name? IMO it’s unnecessary since they’re already in the Qt doc.

Nicolapps:
@j-f1 That’s how the [Qt modules](http://doc.qt.io/qt-5/qtmodules.html) are named. Their name is never used without a leading _Qt_: on the official website, _Qt_ is always repeated. While it can seem odd for a newcomer, it would be more confusing to remove the _Qt_ part from the module names.
Qt 5.11 / Qt Location

Thibaut:
Merged in 67b5e3cb47c80b0f59be30bdd32d5242a2edad6b. Tweaked in e35298c745f09e5657090e01e9810a61a04457ba.

https://devdocs.io/qt~5.11/

Thank you very much for the great PR!

Summary of the changes I made:
- I made the scrapers remove most `<div>`. Those are unnecessary, add weight to pages (saved ~4MB by removing them :P), and often get in the way of the default/consistent styling. By removing them, the CSS override for lined headings becomes unnecessary.
- Remove the icons on the homepage. They required a lot of custom CSS, which makes DevDocs slower for all docs, for little value. Generally I try to have as little custom CSS as possible. That way it's easier to maintain and make style changes across all docs easier + keeps the app fast.
- DevDocs has its own syntax highlighting. To keep things consistent between docs and reduce the need for custom CSS, I always remove the syntax highlighting that comes with the doc. By setting the `data-language` on `<pre>` tags, the code will be highlighted automatically, so long as Prism.js has a highlighter for it (which doesn't appear to be the case for QML unfortunately).
- Increased the max image size a bit because many images were being skipped (from being inlined / made available offline) due to being a bit large than the default limit. 
- Used `frozen_string_literal: true` to try to make the scraper a bit faster.
- Code style changes.

PR #848, opened 2018-07-30
Automatically generate public/docs/docs.json when it's needed

jmerle:
As the title says, this'll automatically generate public/docs/docs.json when it's needed and doesn't exist yet, so that it doesn't have to be in the repository (which it shouldn't, because it only contains user-specific metadata).


j-f1:
Is there any particular reason you put a `.gitignore` file in the`docs` directory instead of just ignoring the directory in the root `.gitignore`?

jmerle:
Nope, didn't think about that yet, but now that you bring it up, that does indeed sound like a better solution. Changed it, everything still works fine.

beaucarnes:
@Thibaut I think this should be merged.

jmerle:
This is no longer needed since Thibaut pushed 54cfc1fbd478cd057186f027ad3d8ab53b3b0cfd.

jmerle:
Nevermind, re-opened, while the file is ignored now it would be prettier to automatically generate it when needed (so we don't keep untracked files around).

PR #851, opened 2018-08-02
Add Handlebars.js documentation

jmerle:
Added a scraper for the Handlebars.js documentation ([Trello card](https://trello.com/c/QSp0MZtz)).

The documentation is licensed under the [MIT license](https://github.com/wycats/handlebars.js/blob/master/LICENSE) (assuming it is under the same license as the project itself).

~I couldn't find a good icon, so I think it's best to use the default icon.~ The icon is licensed under the [Yahoo Inc. BSD license](https://github.com/yahoo/formatjs-site/blob/master/LICENSE.md).

Types:


j-f1:
> I couldn't find a good icon, so I think it's best to use the default icon.

How about using the mustache that’s on [the stickers](https://devswag.com/collections/handlebars/products/handlebars-sticker)? Here’s an [SVG as a `.txt` file to appease GitHub](https://github.com/freeCodeCamp/devdocs/files/2257568/icon.txt)

jmerle:
Thanks for the idea, that SVG looked good but I couldn't transform it into a filled svg (i.e. it would just been a line surrounding the mustache, without the identifiable brown filling it has. I did however search a little deeper and found that Yahoo does have [a website with an SVG version of the Handlebars logo](https://formatjs.io/handlebars/), which I've added in the latest commit (removed the white background and made it square).

Thibaut:
Merged in 8b9c62d7915172eea9904afa3400b64842851d17. Finished in 24abe4c9d914562e16cf16853994243bc3adcdcd.
https://devdocs.io/handlebars/

Great work :)

I changed the `self.type = 'handlebars'` to `self.type = 'simple'`. The way I have a bunch of types listed in `_simple.scss` is actually unnecessary since https://github.com/freeCodeCamp/devdocs/commit/9b28702e3a0dc9d6696ab9446d46132951b7212e when I refactored syntax highlighting. We should change all those to the `simple` type, which will simplify things nicely.

PR #852, opened 2018-08-02
Add Puppeteer documentation

jmerle:
Added a scraper for the Puppeteer documentation ([Trello card](https://trello.com/c/QEZ5vrXD)).

The documentation is licensed under the [Apache License 2.0](https://github.com/GoogleChrome/puppeteer/blob/master/LICENSE).

The logo is also licensed under the [Apache License 2.0](https://github.com/GoogleChromeLabs/pptr.dev/blob/master/LICENSE).

Types:


Thibaut:
Merged in a6b7bd53c2f1f6594850d28c342128e785152802. Finished in 62e07e4e4dd6d672f9fdfe856415e084e7d80e8a.
https://devdocs.io/puppeteer/

Thanks!

I removed the quotes around the event names because they affect the search ranking of those entries. The search primarily relies on the close proximity of a module/function separator like `.` next to the match to rank search results (e.g. rank `foo.test()` ahead of `TestingModule` when you search "test"). A bunch of separators like `#`, `::`, `: `, etc. are normalized into `.` and then the search looks for that character. Quotes are currently not normalized (even though they're not useful for search; I didn't want to change that now since it could have unforeseen consequences, as always with search algo changes), and so its presence was slightly penalizing those entries (by making the event names appear more distant from the separator).

PR #856, opened 2018-08-07
Add Perl alias

zoidbergwill:
Closes https://github.com/freeCodeCamp/devdocs/issues/855

### DIFF
--- a/assets/javascripts/models/entry.coffee
+++ b/assets/javascripts/models/entry.coffee
@@ -70,6 +70,7 @@ class app.models.Entry extends app.Model
     'nginx': 'ngx'
     'numpy': 'np'
     'pandas': 'pd'
+    'perl': 'pl'
     'postgresql': 'pg'
     'python': 'py'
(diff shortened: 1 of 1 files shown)


j-f1:
This alias should already work if you have the Perl docs enabled. Is it not working for you?

zoidbergwill:
Ahhh, sorry about that, I think I was misunderstanding how aliases work, and they just happened to work with my test cases. If it's just replacing a word with another word in searches, then I guess this replacement wouldn't make much difference because of fuzzy matching. Unfortunately most Perl documentation pages don't include `Perl` in the documentation page title, or it'd work how I wanted.

I thought it would only search the docs related to that language, my bad. I misinterpreted the docs.

j-f1:
If you type `pl` and hit <kbd>tab</kbd>, a `Perl` tag should show up in the search box, and searches will be limited to Perl.

Thibaut:
Sounds like this unnecessary. Feel free to reopen if not.

PR #870, opened 2018-09-16
Automatically generate spritesheets

### LINKED ISSUE #787: Sprite file compiler
Currently when a new doc set is added, @Thibaut imports its icon into the sprite file manually, using Photoshop. Instead, we should automate it, using ImageMagick or equivalent.

@Thibaut , can you describe the particular Photoshop plugin or process you currently use? CONTRIBUTING.md only says "Don't modify the icon sprite. I'll do it after your pull request is merged."

jmerle:
This pull request adds a new Thor task that automatically generates the documentation icon spreadsheets. This removes the dependency on Pixelmator, makes it possible to see how new icons look before making a pull request and closes #787.

It works by generating a list containing objects like this:
```ruby
{
  :type => "openjdk",
  :path_16 => "path/to/16.png",
  :path_32 => "path/to/16@2x.png",
  :icon_16 => ChunkyPNG instance,
  :icon_32 => ChunkyPNG instance,
  :row => 0,
  :col => 3,
  :dark_icon_fix => false
}
```

This data is then fed into a method that generates two spritesheets, `docs.png` and `docs@2x.png` using [ChunkyPNG](https://github.com/wvanbergen/chunky_png). The data is also saved into a manifest file, which is read by the `assets/stylesheets/global/_icons.scss.erb` file which generates scss code based off of it. To add erb support to scss files, I added [sprockets-sass](https://github.com/petebrowne/sprockets-sass/) as a dependency.

Whether the dark icon fix needs to be applied is determined based on the contrast of the icon on the dark theme's sidebar background. This is done automatically and based on the rules and formulas from [WCAG 2.0](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#visual-audio-contrast).

All docs without custom icons are displayed with the Mongoose icon, just like it is done at the moment.

The spritesheets are generated at a few moments:
- When someone manually runs the `sprites:generate` Thor task.
- When the Rack server is started in a development or test environment.
- When someone manually runs the `assets:compile` Thor task.

The `log/` directory was added to the gitignore file because it is generated upon running the `assets:compile` task.

The Bluebird 32x32 icon was updated because it was actually a 32x33 icon, and the `sprites:generate` task logs a warning for that (although it does automatically resize the image for the spritesheet).


jmerle:
WIP: It seems like there are some broken tests. I'll fix them after work tomorrow.

j-f1:
Did you add `docs.json` by accident?

jmerle:
Ugh, yes. I'll revert that change after dinner.

jmerle:
Everything's done, this PR is ready to be reviewed and merged.

lib/tasks/sprites.thor
```diff
@@ -0,0 +1,200 @@
...
+      ChunkyPNG::Color.g(color).to_f,
+      ChunkyPNG::Color.b(color).to_f
+    ]
+
+    rgb.map! do |value|
+      value /= 255
+      value < 0.03928 ? value / 12.92 : ((value + 0.055) / 1.055) ** 2.4
+    end
+
+    0.2126 * rgb[0] + 0.7152 * rgb[1] + 0.0722 * rgb[2]
```
kfiles: There are several magic values here. It would be best to replace them with named constants.

jmerle: I've added comments to the pages where the formula is coming from.

PR #872, opened 2018-09-17
Tell Travis to cache bundle

jmerle:
Currently, Travis builds take about 2m15s. About 1 minute of that is spent installing the dependencies. That number can be brought down by telling Travis to cache the dependencies, which should turn that 1 minute into 5 - 10 seconds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: ruby
+
+cache: bundler
+
 before_script:
   - gem update --system
   - gem install bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: ruby
 
-cache:
-  bundler: true
-  directories:
-    - /home/travis/.rvm/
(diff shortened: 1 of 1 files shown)


j-f1:
Can you push an empty commit (`git commit --allow-empty`) to test the cache?

jmerle:
Test results:
- Caching the dependencies makes `bundle install` on Travis run in less than a second as opposed to the current ~1 minute.
- The caching overhead adds about 5 to 10 seconds to the build.
- Caching the `/home/travis/.rvm/` folder in an attempt to reduce the `rvm use` time (which currently takes ~30 seconds) does not work because the cache is loaded *after* the `rvm use` command is ran.

Conclusion: it's useful to add the `cache: bundler` line to the `.travis.yml` file.

Thibaut:
Merged in c22fa4cbf1c01eadbf605f1017496a253c28300a. ❤️

PR #879, opened 2018-10-01
Add Composer

Nicolapps:
This PR adds a scraper for the [Composer documentation](https://getcomposer.org/doc/).

It adds the documentation and additional entries for the configuration options and JSON schema keys, while skipping the FAQ.

Categories:
Composer categories

[Trello Card](https://trello.com/c/kczhwp1k/165-composer)


lib/docs/scrapers/composer.rb
```diff
@@ -0,0 +1,27 @@
+module Docs
+  class Composer < UrlScraper
+    self.name = 'Composer'
+    self.type = 'composer'
```
j-f1: I think you can set this to `'simple'` if you don’t have any special styles.

Nicolapps: Done!

jmerle:
Thanks! Deployed to [https://devdocs.io/composer/](https://devdocs.io/composer/).

PR #883, opened 2018-10-06
Fixed two Python 3.7 types

### LINKED ISSUE #845: Python3 "20.1. email" list number not removed
In Python3(.7) documentation, `20.1. email — An email and MIME handling package` is listed as is instead of being under `Internet Data`

jmerle:
This fixes #845. It groups all email documentation under the `Email` type and all context variable documentation under the `Context Variables` type.

Also fixes a minor issue that caused a lot of errors when generating the documentation because `@levelRegExp` in `lib/docs/filters/python/clean_html.rb` can be `nil`.

### DIFF
--- a/lib/docs/filters/python/clean_html.rb
+++ b/lib/docs/filters/python/clean_html.rb
@@ -22,8 +22,10 @@ module Docs
           end
         end
 
-        css('h2', 'h3', 'h4').each do |node|
-          node.inner_html = node.inner_html.remove @levelRegexp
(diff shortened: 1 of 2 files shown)


Thibaut:
Merged in 31efb9d0f79cd9cc0e23ba45a8926a318dff177f + fixed an additional bug in 0bf7155fbabc3cd0aa87a3caa2e7c4fc4cab4dab (commit message is wrong, bug was in Python 3.7).

PR #884, opened 2018-10-06
Fix internal links to method reference in the Express documentation

### LINKED ISSUE #741: Broken link to express.static inside Express
The link to express.static on https://devdocs.io/express/starter/static-files leads to a page not found error (https://devdocs.io/express/4x/api#express.static) even though the resource does exist (https://devdocs.io/express/index#express.static).

serving static files in express
express static

jmerle:
Fixes #741 by making sure links to the method reference in the Express documentation are replaced with the correct link. Because the root path of the parser is set to `4x/api.html`, the `4x/api.html` file is renamed to `index.html` which breaks all internal links that point towards `4x/api.html`. This pull request fixes that.

### DIFF
--- a/lib/docs/filters/express/clean_html.rb
+++ b/lib/docs/filters/express/clean_html.rb
@@ -42,6 +42,11 @@ module Docs
           node.parent.content = node.parent.content
         end
 
+        # Fix links to the method reference
+        css('a').each do |node|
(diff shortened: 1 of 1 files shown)


Thibaut:
Merged in d1f3322fcc2687fea9c4e794e6d1ea84c135ba8c. Thanks!

PR #885, opened 2018-10-06
Fix Rust entries

### LINKED ISSUE #866: Some elements of Rust std library are missing
I was recently checking Rust `HashMap::retain()` docs via devdocs.io.
To my surprise, it was not there. However, it is present in the original Rust documentation - 
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain

So it seems not all the docs have made it into devdocs.io set.

jmerle:
Fixes the additional entries in Rust's entries filter, which fixes #866. Side-by-side (left is development, right is current):

The issue shows up when there are more than one of these implementation groups (DevDocs only indexes the methods in the first group):

### DIFF
--- a/lib/docs/filters/rust/entries.rb
+++ b/lib/docs/filters/rust/entries.rb
@@ -42,11 +42,14 @@ module Docs
             entries << [node.content, node['id']] unless node.content.include?('Note:')
           end
         else
-          css('#methods + * + div > .method', '#required-methods + div > .method', '#provided-methods + div > .method').map do |node|
-            name = node.at_css('.fnname').content
-            name.prepend "#{self.name}::"
-            [name, node['id']]
-          end
+          css('.method')
+            .select {|node| !node.at_css('.fnname').nil?}
+            .map {|node|
+              name = node.at_css('.fnname').content
+              name.prepend "#{self.name}::"
+              [name, node['id']]
+            }
+            .uniq {|item| item[0]}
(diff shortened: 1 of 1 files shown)


Thibaut:
This adds 12k entries to the doc (before it only had 4k). Is there a way we can reduce the no. of entries a bit?

jmerle:
The only way to reduce the no. of entries would be to leave out certain methods (we could leave out trivial methods like "eq" (`==`), "ne" (`!=`) and "cmp" (`<`/`<=`/`>=`/`>`)), but then you'd have to accept that the entries list is not complete.

Correct me if I'm wrong, but with 12k entries Rust would not even be close to the biggest docs, so I don't think it's a problem. I think it's more of a problem that we're missing 8k entries now.

Thibaut:
Right, I was thinking more along the lines of things like inherited functions or functions included from other modules, which can create many duplicates without much value. Overall DevDocs's search is better the fewer "noisy" entries there are. The goal isn't to try to keep the no. of entries below a certain fixed number; it's more to try and have as few "useless" entries as possible, to maximize the usefulness/relevancy of search results.

Thibaut:
Merged in 3dbc6052bf131fa37516d473e44732ed7e2ad1a2 and made some additional improvements in 7b7aa34b7069a89cfc17b7e45956e5e84b3c1349.

PR #890, opened 2018-10-09
Fix CSS variable support check

jmerle:
The check to see if the browser supports CSS variables broke DevDocs Desktop, as reported in egoist/devdocs-desktop#85. I'm not sure exactly why it doesn't work correctly, but by using `CSS.supports('--t', '0')` instead of `CSS.supports('--t', '0')` the issue is fixed. This newer way was tested to return `true` on Chrome 59 and 69 (which both support CSS variables), and on Chrome 47 (which doesn't support CSS variables) it returns `false`.

Chrome 47:

Chrome 59:

Chrome 69:

### DIFF
--- a/assets/javascripts/app/app.coffee
+++ b/assets/javascripts/app/app.coffee
@@ -245,7 +245,7 @@
         matchMedia:         !!window.matchMedia
         insertAdjacentHTML: !!document.body.insertAdjacentHTML
         defaultPrevented:     document.createEvent('CustomEvent').defaultPrevented is false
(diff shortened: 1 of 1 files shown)


Thibaut:
Thanks! Just came to the same conclusion in `847cfbb682c3bccdfe9e53c1d72a541787d1619f` 😄 

I'll get this out as soon as I get home.

Thibaut:
Just merged and deployed this (05aaf5eeddc447dce4321109810b3eed81f16354), but it broke Safari now…

Thibaut:
33c986fcf0a7773771e19237337d55cabf217757

PR #893, opened 2018-10-09
Fixing dark mode external link display

iaian:
- Only apply %external-link in a:after

Fixes issue : https://github.com/freeCodeCamp/devdocs/issues/892

### DIFF
--- a/assets/stylesheets/components/_page.scss
+++ b/assets/stylesheets/components/_page.scss
@@ -11,7 +11,9 @@
   > h1 { @extend ._lined-heading; }
   > h1:first-child { margin-top: 0; }
 
-  a[href^="http:"], a[href^="https:"] { @extend %external-link; }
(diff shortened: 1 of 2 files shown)


Thibaut:
This is likely breaking the non-dark mode. The `%external-link` Sass class already applies to the `&:after` pseudo element:

https://github.com/freeCodeCamp/devdocs/blob/33c986fcf0a7773771e19237337d55cabf217757/assets/stylesheets/global/_classes.scss#L140-L154

I think the issue is with this:

https://github.com/freeCodeCamp/devdocs/blob/33c986fcf0a7773771e19237337d55cabf217757/assets/stylesheets/global/_classes.scss#L155-L157

it should be `html._theme-dark %external-link &:after`.

Thibaut:
Fixed in d72de330aa4086297095f4c7018c19263d07335e.

PR #900, opened 2018-10-17
delete whitespace

nhungltm:


j-f1:
This space is necessary for the output to be formatted correctly.

PR #901, opened 2018-10-17
change a to link_to

nhungltm:


j-f1:
These files aren’t processed by the server before they’re transmitted, and even if they were, I don’t think our server has the `link_to` helper.

PR #903, opened 2018-10-17
change condition clearly

nhungltm:


j-f1:
This inserts extra unnecessary whitespace in the production output.

PR #904, opened 2018-10-17
change control structure clearlier

nhungltm:


j-f1:
CC @Thibaut — is this style intentional?

Thibaut:
Yeah it was, so that the output is nicely indented (check the source of devdocs.io). Not really needed, but I don't see a need to change it.

PR #918, opened 2018-10-26
Update in media Query

atir11122:


j-f1:
Why the change?

Thibaut:
This is for the 404 page. Doesn't seem to be of any use.

PR #921, opened 2018-10-27
Started a small step for devdocs explanation

jstodd42:


j-f1:
This seems to be already covered by the homepage.

PR #928, opened 2018-11-06
Add GnuCOBOL documentation

jmerle:
Added a scraper for the GnuCOBOL documentation.

The documentation is licensed under the [GNU Free Documentation License](https://open-cobol.sourceforge.io/HTML/gnucobpg.html#Appendix-C-_002d-GNU-Free-Documentation-License).

GnuCOBOL doesn't have an official logo, so the default icon should be used.


lib/docs/filters/gnu_cobol/clean_html.rb
```diff
@@ -0,0 +1,56 @@
...
+        css('div[align="center"]').each do |node|
+          if node.content.include?('=' * 50)
+            node.replace('<hr>')
+          else
+            node.remove_attribute('align')
+            node.name = 'h4'
+          end
+        end
+
+        # Remove all hr's after h4's
```
j-f1: Can you also remove all hrs before h3s?

j-f1: Can you also remove all hrs before h3s?

jmerle:
Implemented everything you suggested. As for the x.y. and x.y.z. headers, they now look like this:

lib/docs/filters/gnu_cobol/clean_html.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        # Remove everything after Appendix B
+        # This includes the license text, the document changelog, the compiler changelog and the footnote
+        start_element = at_css('a[name="Appendix-C-_002d-GNU-Free-Documentation-License"]').previous_element
+        next_element = start_element.next_element
+        until start_element.nil?
+          start_element.remove
+          start_element = next_element
+          next_element = start_element.nil? ? nil : start_element.next_element
+        end
```
j-f1: This seems to remove the content of Appendix B, too. However, appendix B is just a list of all the reserved words, so it might be reasonable to remove it too.

jmerle:
I've added the logo you suggested. I considered adding it before but thought it wouldn't look good because of the details in it. Now that it's actually possible to generate the spritesheet programmatically I noticed that it wasn't so bad after all.

As for the `acronym` thing, all `acronym` tags are now converted to `span` tags so the underlining is removed.

PR #935, opened 2018-11-12
update TensorFlow version

suyash:


Thibaut:
I don't think this works, because the source site has changed since the last scrape. It'll require more work.

Per the guidelines: https://github.com/freeCodeCamp/devdocs/blob/master/.github/CONTRIBUTING.md#updating-existing-documentations

> Please don't submit a pull request updating the version number of a documentation, unless a change is required in the scraper and you've verified that it works.

Issue #938, opened 2018-11-23
option to remove single-documentation search parameter without emptying search box


jomel:
![image](https://user-images.githubusercontent.com/3426561/48942154-268e9e80-ef16-11e8-96c3-7adf6eedf915.png)

If I am currently searching a single documentation (eg by clicking through from vscode addon or by using keyword+tab) according to the Guide I have to remove the search params in the box & click backspace to remove single-documentation filter.

This is really annoying. I end us cutting my word, clicking backspace and pasting again. or opening a new devdocs.io tab!

some suggestions:
- when the cursor in the search box is at the front (between the grey documentation "brick" and the search words) backspace should remove the documentation, but keep the search words
- clicking on the documentation "brick" in the search bar should remove it.
- the vscode extension should have an option whether you want the "language filer" automatically applied or not.

jmerle:
I agree on your first two suggestions. I'm actually amazed that your first suggestion isn't already the current behavior. Will see if I can make time tonight.

As for the third suggestion, you'll need to ask the maintainer of the vscode extension if he can implement it.

jmerle:
Implemented in #942 which has been merged and deployed.

PR #942, opened 2018-11-25
Remove scope by clicking on it or by backspacing into it

jmerle:
Adds two improvements to the search scope as suggested in #938:
- When a scope is selected and the backspace is pressed when the insertion point is all the way to the left, the scope is removed:

- When a scope is selected and you press on the search tag, it is removed:

### DIFF
--- a/assets/javascripts/views/search/search.coffee
+++ b/assets/javascripts/views/search/search.coffee
@@ -30,6 +30,9 @@ class app.views.Search extends app.View
       .on 'results', @onResults
       .on 'end', @onEnd
 
+    @scope
+      .on 'change', @onScopeChange
+
     app.on 'ready', @onReady
     $.on window, 'hashchange', @searchUrl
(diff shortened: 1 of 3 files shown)


assets/javascripts/views/search/search.coffee
```diff
@@ -138,6 +141,10 @@ class app.views.Search extends app.View
     $.stopEvent(event)
     return
 
+  onScopeChange: =>
+    @value = ''
```
Thibaut: Took me a while to figure out why we need this. Instead of doing `$.trigger @input, 'input'` below I would move all the logic for "re-trigger search after scope change" in here so it's a bit more clear:

```coffee
onScopeChange: =>
  @value = '' # reset memoized value to re-run the current search (with the new scope)
  @onInput()
```

j-f1:
Can you change the cursor on the tag to make it more clear that it’s clickable, especially since it wasn’t in the past?

PR #952, opened 2018-12-13
Add pony

haarts:
This is the documentation for [Pony](ponylang.io).

There's an Entry called 'Package' with almost all, well, package. It is sort of an intro, very useful. But as the entries are alphabetically sorted this ends up at random places. Is there a way to have this entry always on top or something?


lib/docs/scrapers/pony.rb
```diff
@@ -0,0 +1,12 @@
+module Docs
+  class Pony < UrlScraper
+    self.type = 'pony'
+    self.release = '0.25.0'
+    self.base_url = 'https://stdlib.ponylang.io/'
+
+    html_filters.push 'pony/container', 'pony/clean_html', 'pony/entries'
+    options[:attribution] = "Me"
```
j-f1: Can you update this to include both the copyright and license information for the docs?

lib/docs/filters/pony/clean_html.rb
```diff
@@ -0,0 +1,9 @@
+module Docs
+  class Pony
+    class CleanHtmlFilter < Filter
+      def call
+        doc
+      end
+    end
+  end
+end
```
j-f1: If you don’t need to do anything here, you can remove this filter entirely.

j-f1:
If you want the intro page to be the homepage, you’d need to set the initial path to an appropriate value so it would show up.

cc @Thibaut — is it ok that this project doesn’t have 5k stars yet?

lib/docs/scrapers/pony.rb
haarts: Ow, how very sloppy of me. Will fix of course.

lib/docs/filters/pony/clean_html.rb
haarts: Ah, I was under the impression it was mandatory. I'll remove it.

Thibaut:
>  is it ok that this project doesn’t have 5k stars yet?

Unless the project is very new or small, I don't think that's a problem. We have scrapers for docs that don't meet the criteria.

haarts:
I see that 'Changes requested' but as far as I can see I've done what was requested?

haarts:
*bump*. When merged I can remove it from my mental checklist. :)

jmerle:
Don't worry, it will be merged some time, but currently there is no active maintainer with permissions to merge and deploy, so it might take a while longer. We are currently waiting for freeCodeCamp to get @ahmadabdolsaheb, @j-f1 and me the correct permissions, but it's unknown when that'll happen.

jmerle:
Thanks! Deployed to [https://devdocs.io/pony/](https://devdocs.io/pony/). You can get this off of your mental checklist now ;)

I prefixed all package entries with "1. " to ensure they show up on top. I also added entries for all functions and fields and updated the generated doc's styles to align more closely with the rest of the docs on DevDocs.

PR #955, opened 2018-12-17
Update Qt docs

tronical:
* Added Qt 5.12 docs (latest LTS release)
* Use a stable URL for Qt 5.11 docs


tronical:
Thanks for the review :). What can I do to get this change landed?

PR #961, opened 2019-01-03
Update PHPUnit documentation (7.4)

cdtweb:


jmerle:
Nevermind, my review was wrong. This won't work since PHPUnit moved to readthedocs.io, so our scraper and filters will need to be rewritten to update the PHPUnit docs to support versions 7 and 8 (and drop support for scraping versions 4, 5 and 6).

I'm pretty sure we can write a generic scraper for docs on readthedocs.io to make it easier to scrape those docs for DevDocs, but that's outside the scope of this PR so I'm closing this.

PR #970, opened 2019-01-26
Add ReactiveX documentation

jmerle:
Added a scraper for the ReactiveX documentation.

The documentation is licensed under the [Apache License 2.0](https://github.com/ReactiveX/reactivex.github.io/blob/develop/LICENSE).

The official icon is also licensed under the [Apache License 2.0](https://github.com/ReactiveX/reactivex.github.io/blob/develop/LICENSE). Because I could only find a 32x32 version with a transparent background in the ReactiveX repositories, I used Imagick's `convert` to create a 16x16 version of it.

Image downloading is disabled because the docs contain a lot of images which makes scraping really, really slow (10+ seconds for some pages), and because enabling it causes 14 pages to be skipped because of an HTTP error. Not sure why that is, but when `options[:download_images]` is set to `false`, it works fine.

Types:


j-f1:
Looks like you’ve got a merge conflict on Prism.js.

jmerle:
Fixed :)

lib/docs/filters/reactivex/clean_html.rb
```diff
@@ -0,0 +1,65 @@
...
+        css('span').each do |node|
+          next unless node.content == 'TBD'
+          node.xpath('./ancestor::div[contains(@class, "panel-default")][1]').remove
+        end
+
+        # Remove the : at the end of "Language-Specific Information:"
+        css('h2').each do |node|
+          node.inner_html = node.inner_html.gsub('Information:', 'Information')
+        end
+
```
j-f1: Since this doc uses h1s in its content, all the headers should be taken down one level since the h1 is only used for the title of the page in DevDocs.

j-f1: On http://localhost:9292/reactivex/documentation/operators#categorized, the linked code blocks don’t show up properly. This is because the HTML structure is like `<a><strong><code>...</code></strong></a>`, so the `<code>` element doesn’t know it’s supposed to inherit its color. This could be fixed either by stripping out the extra `<strong>`, moving it outside of the `<a>`, or changing the base DevDocs CSS to match a `<code>` anywhere inside a `<a>`. Here’s the offending rule:

https://github.com/freeCodeCamp/devdocs/blob/6efad323ff55edd81e72a699f55ae5e4d7c80c73/assets/stylesheets/global/_base.scss#L121

assets/stylesheets/pages/_reactivex.scss
```diff
@@ -0,0 +1,11 @@
...
+  @extend %simple;
+
+  img {
+    max-width: 50%;
+  }
+
+  figure {
+    margin: 0;
+  }
+}
```
j-f1: On http://localhost:9292/reactivex/documentation/operators#tree, the tree isn’t formatted the same as it is [on the original site](http://reactivex.io/documentation/operators.html#tree), making it a little harder to understand and much larger vertically than the original.

j-f1: On http://localhost:9292/reactivex/documentation/observable, `<dfn>` elements are rendered with a “?” cursor and a dotted underline, implying that you can hover it to see a definition. On the original site, they’re rendered as plain *italics*.
```suggestion
  dfn {
    cursor: text;
    text-style: italic;
    text-decoration: none;
  }
}
```

lib/docs/filters/reactivex/clean_html.rb
j-f1: On the operators pages, there are h3s at the top of each page with a description of the operator. Can you change this element to either be a normal `<p>` or a `<blockquote>` (or something else that isn’t a header)?

jmerle: Changed it to a `<blockquote>`:

jmerle: I've updated it so that all h1's except for the first are changed into `h2`'s. There are only around 3 pages with multiple h1's and most other pages don't look well if you take all headers down by one level, so this is probably the best solution.

jmerle:
Processed your review comments. Regarding the Backpressure Operators page, it is now properly capitalized and categorized under "Manual".

lib/docs/filters/reactivex/clean_html.rb
j-f1: Looking  at this again, the `<strong>` seems to have some significance in this doc, so going the route of changing the above selector would be better, so as to retain the boldface. However, it seems there’s an additional rule specifying `font-weight: normal` for code spans, so that needs to be either overridden or removed (if it’s not necessary).

assets/stylesheets/pages/_reactivex.scss
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    dt {
+      float: left;
+      clear: left;
+
+      margin-top: 0;
+
+      &:before {
+        content: "…";
```
j-f1:
```suggestion
        content: "\2026";
```

Seems Rails doesn’t like non-ASCII content in assets.

PR #972, opened 2019-02-01
Dynamic favicon

### LINKED ISSUE #803: Set favicon by documentation set
Currently, all DevDocs tabs have the same generic favicon, making them hard to tell apart.

It would be useful to set favicon by documentation set, so that Python documentation is immediately distinct from Ansible documentation, etc.

jmerle:
This PR makes the favicon change depending on the documentation that is shown. Closes #803.


j-f1:
What do you think about overlaying the docset icon on the DevDocs icon so tabs can still be picked out quickly?

image

<sup>Based on [GitHub’s parent-child avatar style](https://styleguide.github.com/primer/components/avatars/#parent-child-avatars)

jmerle:
Good call, that parent-child style does make it obvious that it's DevDocs' JS documentation and not something like MDN or another source. I will give it a try this weekend.

jmerle:
The latest commit makes it look like this:

ludvigeriksson:
Will this be merged? It looks great and I would love to have it.

j-f1:
Yep! We’re having a meeting soon to work out maintenance issues but after that we’re hoping to get back to merging PRs here.

jmerle:
Just tested it again, merging #870 into master did not break this. Ready to be reviewed.

yurikhan:
Looks like there might be some issue on HiDPI displays?

When I open a page in Ansible, I get the Angular icon. When I open CSS, I get the C++ icon. When I open C++, I get half the Angular icon and half the C++ icon.

image

I’m on Firefox 68/GNU/Linux if that matters.

jmerle:
@yurikhan I pushed an update that should fix the issue.

yurikhan:
Works now, thank you!

j-f1:
I’m getting an error on macOS Safari where `paddingSize` is non-integer so the call to `new ImageData` fails with `IndexSizeError: Length is not a multiple of sw`

Issue #971, opened 2019-02-01
Analytics Opt-Out


ghost:
![screenshot_20190201_115701](https://user-images.githubusercontent.com/440298/52101995-0ca9c000-2619-11e9-86d5-2ce8fa3ebaad.jpg)

ice cat android emui 9

j-f1:
Do you have cookies enabled?

j-f1:
DevDocs needs cookies to store the list of docs you’ve enabled.

j-f1:
I’m not seeing those cookies — are you sure those aren’t coming from an extension? I’ve got a CloudFlare cookie that’s scoped to DevDocs.io, a Google Analytics cookie, some Gauges cookies, and various settings I’ve set in DevDocs.

jhabdas:
Looks like the above snap was coming from the Ledger Manager extension, and activated _specifically_ when visiting DevDocs.io (and not other domains I checked). I have since removed the extension, but I still have a copy if you want to analyze it to determine what it may have been doing to your users.

Here are the cookies coming from your origin, which also include a CloudFlare tracker I'd like to get rid of while still enjoying DevDocs from an IP address in the EU:

<img width="550" alt="screen shot 2019-02-04 at 7 52 00 pm" src="https://user-images.githubusercontent.com/440298/52206889-97f3b180-28b6-11e9-9cbd-8b48dfc4452e.png">

Apart from the trackers the majority of the UX data should probably be moved into Web Storage and the cookies made optional thereafter unless you pop a consent disclaimer.

j-f1:
I believe we store the data in cookies so the server can render the pages and Application Cache manifest properly. If you’re worried about tracking, I’d suggest the DuckDuckGo extension.

jhabdas:
I know this isn't the sexiest subject, but it's been almost a year since GDPR rolled out and I've never seen a cookie disclaimer on DevDocs.io. Perhaps it was overlooked or you're blocking IPs in EU?

https://eugdpr.org/the-regulation/gdpr-faqs/

> **What constitutes personal data?**
> The GDPR applies to ‘personal data’, meaning any information relating to an identifiable person who can be directly or indirectly identified in particular by reference to an identifier. This definition provides for a wide range of personal identifiers to constitute personal data, including name, identification number, location data or online identifier, reflecting changes in technology and the way organisations collect information about people.

[...]

PR #975, opened 2019-02-11
Stop using an iframe for the GitHub star button

j-f1:
Ref #971.


assets/javascripts/templates/pages/root_tmpl.coffee.erb
```diff
@@ -39,7 +39,7 @@ app.templates.intro = """
       <li>DevDocs works <a href="/offline">offline</a>, on mobile, and can be installed on <a href="https://chrome.google.com/webstore/detail/devdocs/mnfehgbmkapmjnhcnbodoamcioleeooe">Chrome</a>.
       <li>For the latest news, follow <a href="https://twitter.com/DevDocs">@DevDocs</a>.
       <li>DevDocs is free and <a href="https://github.com/freeCodeCamp/devdocs">open source</a>.
-          <iframe class="_github-btn" src="//ghbtns.com/github-btn.html?user=freeCodeCamp&repo=devdocs&type=watch&count=true" allowtransparency="true" frameborder="0" scrolling="0" width="100" height="20"></iframe>
```
jhabdas: Another idea, not sure if you tried it, would be to add the sandbox attribute.

j-f1: IMO a static SVG is better anyway, so there’s no point in using an iframe.

PR #979, opened 2019-02-12
Cookie consent

### LINKED ISSUE #971: Analytics Opt-Out

jmerle:
This PR should make DevDocs compliant with the GDPR and the upcoming ePrivacy regulation. This closes #971.

The following has been done:
- DevDocs now asks for permission before enabling Google Analytics and Gauges.
- The user is only asked for consent once per browser session until he/she either accepts or declines.
- The user can always change his decision in the settings. If the user disables tracking cookies, all existing cookies are deleted immediately.
- Since the default is no tracking cookies, all existing analytics cookies will be deleted once this PR is merged into production and all current users will have to either give consent or decline it.
- The privacy policy has been updated to reflect the above.

This PR does not disable the `__cfduid` cookie. Although there used to be a comment in #971 that (quite heavily) opposed this cookie, it seems like that is now deleted. I also do not agree with that now-deleted comment, since CloudFlare [explicitly states that the cookie does not correspond to any user id in the application nor does it store personally identifiable information](https://support.cloudflare.com/hc/en-us/articles/200170156-What-does-the-Cloudflare-cfduid-cookie-do-).

Consent notification:

Settings option:

Privacy policy:


assets/javascripts/tracking.js
```diff
@@ -1,28 +1,32 @@
...
-    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-    ga('create', 'UA-5544833-12', 'devdocs.io');
-    page.track(function() {
-      ga('send', 'pageview', {
-        page: location.pathname + location.search + location.hash,
-        dimension1: app.router.context && app.router.context.doc && app.router.context.doc.slug_without_version
+  if (app.config.env === 'development') {
```
j-f1: Should the change in this line be reverted?

j-f1:
Would it be possible to keep the cookie so stats remain accurate if the user decides to consent to tracking? The cookies could still be removed if the user doesn’t explicitly consent.

assets/javascripts/tracking.js
jmerle: You're right, I forgot to revert that change. Thanks!

jmerle:
That would make our tracking cookies opt-out to current users instead of opt-in, which doesn't quite sound right to me.

Also, after this change we can expect the "unique user count" statistics to be skewed anyways, so I don't really see the problem in re-creating the cookies for users who accept our tracking cookies.

PR #981, opened 2019-02-25
Bump homebrew to 2.0.2 (latest version)

szb512:
[ x] Tested the scraper on a local copy of DevDocs
[x ] Ensured that the docs are styled similarly to other docs on DevDocs

This will pretty much bump homebrew to 2.0.2.

### DIFF
--- a/18.06
+++ b/18.06
@@ -1,0 +1,45 @@
+  Docker, Inc. and other parties may also have trademark rights in other terms used herein.
+    HTML
+
+    version '18' do
+      self.release = '18.06'
+      self.base_url = 'https://docs.docker.com/'
+
+      html_filters.push 'docker/entries', 'docker/clean_html'
+
+      options[:container] = '.wrapper .container-fluid .row'
+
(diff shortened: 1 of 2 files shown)


szb512:
Did I forget to mention that the process uses homebrew to upgrade?

szb512:
Also it will move docker from 18.03 to 18.06.

j-f1:
Please upgrade the original source files instead of creating new ones.

szb512:
What files do I need to edit then?

jmerle:
Requesting @j-f1's opinion on this. I don't think we need new PR's for minor and patch releases when there are already open PR's for the major versions (Homebrew 2.0.2 at #978 and Docker 18.03 at #976) and when the documentation structure hasn't changed. Most PR's will already be outdated by the time we can get them reviewed, so I think it is our responsibility as maintainer to ensure we update to the latest version when we merge a PR that updates a scraper. For that reason I'd vote to close this PR.

As to answer your question, you can look in the PR's I linked above to see the changes the contributors made.

PR #986, opened 2019-03-10
Automatically check for documentation updates

jmerle:
This PR adds the `updates:check` Thor command which automatically checks for outdated documentations. This removes the need for the ["Request updates here" Trello card](https://trello.com/c/2B0hmW7M) and should make it easier to keep documentations up-to-date.

The command can check either all docs or the docs given to it with arguments (i.e. `thor updates:check ruby minitest nginx`). After it's done checking all versions, it will show the results divided in three categories: outdated docs, up-to-date docs and docs that could not be checked (which can happen if a scraper does not override `get_latest_version`).

In order to check the latest version of documentations, a few methods have been added to `Docs::Doc`, which all scrapers inherit from (the scraper reference has also been updated with these new methods):
- `get_scraper_version` returns either the latest `self.release` value or the Epoch time of when scraped docs were last pushed to DevDocs production if `self.release` is not set.
- `get_latest_version` returns a `NotImplementedError` by default and should be overridden by all parsers. If `self.release` is set, it should return the latest version of the documentation that is being scraped. If `self.release` is not set, it should return the Epoch time of when the documentation that is being scraped was last updated. If the documentation that is being scraped will never change, this should return `1.0.0`.
- `is_outdated` compares the scraper version with the latest version to decide whether the documentation is outdated. This method ignores patch updates.
- `fetch`, `fetch_doc`, `fetch_html`, `get_npm_release`, `get_latest_github_release`, `get_github_tags` and `get_github_file_contents` which are utility methods which are used in `get_latest_version` implementations.

I have already implemented `get_latest_version` for the 146 parsers we currently have. I will also implement it (or help to implement it) on parsers that are added before this PR is merged.

The command accepts three options:
- `--github-token` can be used to authorize any calls to the GitHub API with a Personal Access Token, after which the 60 requests per hour limit no longer applies.
- `--verbose` can be used to enable extra logging (like which url's are being fetched).
- `--upload` can be used to upload the results to a new GitHub issue.

When `--upload` is given as option, a few things are done after checking all versions:
1. The command checks whether `--github-token` is set and whether it belongs to @devdocs-bot (which is the only account that should be uploading results).
2. A new issue is created in this repository called "Documentation versions report for [month] [year]" with lists of outdated docs, up-to-date docs and docs that could not be checked.
3. If a previous issue exists from the previous month, a comment is placed noting it was superseded by the new issue and the previous issue is closed (if it was not already closed by a maintainer).

An example of the issue that will be created: [jmerle/devdocs#16](https://github.com/jmerle/devdocs/issues/24)

To automate this process, a new cron job should be created in DevDocs' Travis CI configuration after this PR is merged. It should run on the `master` branch with interval set to `monthly` and options set to `Always run`. I have already added the `GH_TOKEN` environment variable to the Travis CI configuration containing a Personal Access Token for the @devdocs-bot user with the `public_repo` scope. After this is done a new issue will be created every month containing a list of outdated documentations, which maintainers can use to ensure DevDocs stays up-to-date.


j-f1:
Would it make sense to pull the latest version from GitHub for some of these scrapers by fetching `https://github.com/:user/:repo/releases/latest` and pulling the version number from the redirected URL?

jmerle:
I changed it to something like that. The old behavior was that `get_latest_github_release` would simply return the object described [here](https://developer.github.com/v3/repos/releases/#get-the-latest-release). Now it returns the tag name of the latest release (the thing you see in the url after being redirected) and it automatically removes the "v" preceding the version if there is one.

j-f1:
Why do some scrapers pull from the latest GitHub release while others pull from the docs website?

jmerle:
I believe my reasoning for that at the time was that sometimes the docs website only shows major versions (like "3" or "4") while the GitHub releases would also show minor and patch versions (like "3.1.2" or "4.3.0") in which the docs would also get updated. Although the command does not report patch updates as outdated documentations, it does do so with minor updates which is why that information was important.

While implementing the `get_latest_version` method on the scrapers the first thing I would try was fetching the version from the docs website, but if that was not possible due to the above or for another reason (like the versions being loaded asynchronously for example or for it being unnecessarily hard) I would instead use the latest release on GitHub/NPM/Packagist.

PR #990, opened 2019-03-20
Add support for Clojure 1.10 and "master"

allentiak:

- [X] Tested the scraper on a local copy of DevDocs

Hi, @j-f1, @Thibaut, @jmerle  
I'm having problems making this doc update work. Could anyone of you please help me on this?

I have followed the pattern of previous Clojure doc updates (such as https://github.com/freeCodeCamp/devdocs/commit/6b33e3c9d62cdac90c3b149f6b2b0e8ece171305). However, my change does not work...

When testing it locally, I get a lovely 404...

```
-> % bundle exec thor docs:download clojure@1.10
(1/1) Clojure 1.10 FAILED (OpenURI::HTTPError: 404 Not Found)
Done
```

The strange about this is that if, instead of adding a new entry, I simply replace the version string '1.9' with '1.10', I get the same 404 error when executing `bundle exec thor docs:download clojure@1.10`...

Could anyone help me on this, please?

(I have requested this PR because this doc version has already been requested many times in the Trello board during the last three months.)


j-f1:
The `docs:download` command attempts to download the pre-scraped docs directly from the DevDocs server. I think you want the `docs:generate` command, which will scrape the source site and create the new documentation.

allentiak:
Thanks, @j-f1! 
It is as you said.

I will check my code and update this PR's title when ready.

Thanks again!
--Leandro

allentiak:
@j-f1 I think it's ready.
(I also added support for the latest "development" branch. What do you think about this?)

j-f1:
LGTM, but we’re in a maintenance situation at the moment where we can’t merge changes to the code. Hopefully we’ll have this resolved soon!

allentiak:
Thanks for your quick reply! Looking forward to the completion of the
rewrite!

BTW: Maybe you could express that the code is in maintenance mode in the
README, so others PR authors will not feel neglected?
I see there are many pending PRs...

>

jmerle:
Hi!

Looks good, although I did remove the master/development version. We generally don't include those because the content inside them can quickly become outdated and the latest stable version is usually "good enough".

allentiak:
Thank you so much, @jmerle!

Issue #985, opened 2019-03-06
Ramda documentation missing entries


colepeters:
Thanks for all the work on this awesome resource!

I've noticed that the Ramda documentation is missing some entries, namely [`mergeRight`](https://ramdajs.com/docs/#mergeRight) and [`mergeLeft`](https://ramdajs.com/docs/#mergeLeft), which are available in the Ramda docs. (There may be others missing, but these are the only two I've spotted so far.) Maybe the docs need to be re-scraped?

Tzolkin:
In the version 0.26.1 some cards don't have the 'small' ```CSS``` and when you try to re-scraped the docs throw the error:
```
ERROR:
  http://ramdajs.com/0.26.1/docs/
  NoMethodError: undefined method `parent=' for nil:NilClass

  /devdocs/lib/docs/filters/ramda/clean_html.rb:14:in `block in call'
  /devdocs/lib/docs/filters/ramda/clean_html.rb:11:in `call'
  /devdocs/lib/docs/core/instrumentable.rb:15:in `instrument'
  ...
  /devdocs/lib/docs/core/doc.rb:107:in `store_pages'
  /devdocs/lib/docs.rb:100:in `generate'
  /devdocs/lib/tasks/docs.thor:300:in `generate_doc'
  /devdocs/lib/tasks/docs.thor:106:in `generate'
```

[Adding a guard](https://github.com/Tzolkin/devdocs/blob/feature/ramda/lib/docs/filters/ramda/clean_html.rb#L14) clause we can fix the issue, with results:
```
Entries: (16 KB, +1 KB)
  Updated: 245
  Created: 10
    + R.composeWith
    + R.hasPath
    + R.includes
    + R.mergeLeft
    + R.mergeRight
    + R.move
    + R.otherwise
    + R.pipeWith
    + R.then
    + R.thunkify
  Deleted: 1
    - R.memoize
Types:
  Updated: 8
  Created: 0
  Deleted: 0
Files: (0.3 MB, +0.1 MB)
  Updated: 1
  Created: 0
  Deleted: 0
Done
```
<img width="1680" alt="Screen Shot 2019-04-16 at 21 30 54" src="https://user-images.githubusercontent.com/1302551/56256581-ef6c8580-608e-11e9-814c-6d205b29f3a7.png">

colepeters:
@Tzolkin That's great! Are you able to file a PR since you've got this working?

Tzolkin:
[...]

PR #999, opened 2019-04-24
Fix Ramda scraper

Tzolkin:
Fixed and issue when re-scraped the docs in Ramda documentation version 0.26.1
This include a new version for Ramda documentation.


colepeters:
Thanks so much, looking forward to this!
(This issue raised in #985)

GU5TAF:
The current docs for ramda are two versions behind since the scraper stopped working, would be great to get it up to date.

Perhaps this PR should simply be renamed to "Fixed ramda scraper" so it's a little more clear how important the fix is?

PR #1011, opened 2019-05-26
Fix layout preferences on Firefox

jmerle:
Currently layout preferences are loaded based on the cookies sent in the HTTP request to devdocs.io. This does not work on recent versions of Firefox, as has been reported in #992.

After this PR is merged, layout preferences are also loaded locally based on the user's stored cookies. By loading preferences client-side, we no longer need to rely on the optional [`Cookie`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) header. This fixes #992.

Locally loading layout preferences happens before the content is shown, so there is no flickering between light and dark themes on slow connections. Performance does not seem to be affected neither.

The code to load the layout preferences based on the cookies sent with the HTTP request has not been removed as there seem to be some dependencies between that code and the AppCache-related code.

Issue #774, opened 2018-03-25
Switch to a Service Worker


j-f1:
[It looks like](https://jakearchibald.github.io/isserviceworkerready/) Service Workers are supported on the latest versions of all of the browsers.

Some ideas:

- automatically cache anything the user looks at while online
- use [Web Push](https://devdocs.io/browser_support_tables/push-api) to update docsets in the background even if DevDocs isn’t open (maybe if they disable auto-updates, we could prompt them with a desktop notification)
- preload linked pages when the user tabs to a link to them, or hovers over the link
- preload the top few search results for instant loading

If we use Service Workers, we won’t have to do anything different when offline vs online — just fetch the docset from the URL, and the worker will proxy it to the cache, even if the user is online (we can also fetch an updated version if the cached copy is out-of-date and store it for the next time the page gets loaded).

Demo will be available [here](https://fortunate-jumper.glitch.me) at some point (it’s still on AppCache right now)

cglong:
It looks like, [as of EdgeHTML 17.17063](https://blogs.windows.com/msedgedev/2018/02/06/welcoming-progressive-web-apps-edge-windows-10/#mBXG9JCltyA7IOQ7.97), Service Workers are enabled by default 😄

It'd be good to support this, as Service Workers are [a requirement for automatic indexing into the Microsoft Store](https://docs.microsoft.com/en-us/microsoft-edge/progressive-web-apps/microsoft-store#automatic-pwa-importing-with-bing).

j-f1:
Updated. Thanks!

jmerle:
[Since Chrome 68](https://developers.google.com/web/updates/2018/07/nic68#a2hs) the "Add to homescreen button" is no longer available for manual use on Android. Instead, websites can trigger an "Add to homescreen" dialog if the website meets certain [criteria](https://developers.google.com/web/fundamentals/app-install-banners/#criteria). One of these criteria is registering a service worker with a `fetch` event handler, so that's another vote towards this issue.

ryanburnette:
Thumbs up to converting devdocs from appcache to serviceworkers.

eloyesp:
I would vote against that, because I like this project and [Service Workers might be insecure][1] so are disabled by many people (me included).

[1]: https://security.stackexchange.com/questions/202197/privacy-tracking-concerns-regarding-service-workers

ryanburnette:
[...]

PR #1022, opened 2019-07-11
Replace the App Cache with a Service Worker

### LINKED ISSUE #557: Can we make devdocs a chrome extension instead of a chrome app?
Can we make devdocs a chrome extension instead of a chrome app?
Based on this article, google shall remove chrome apps in the next few months...
http://www.theverge.com/2016/8/19/12555052/google-shutting-down-chrome-apps
Also having a visual indicator before going offline on the extensions pane is a nice reminder to (Feature:) click on it and let it re-install documentation for offline use.

jmerle:
This PR contains the following changes:
- Replaced the App Cache with a Service Worker. Everything still functions the same, we just don't rely on the deprecated App Cache anymore. This change bumps the browser version requirements for Edge from 16+ to 17+, for Safari from 9.1+ to 11.1+ and for iOS from 10+ to 11.3+. All of these were released in the beginning of 2018, so this should be fine under our "we support recent versions of modern browsers" policy. This partially implements #774.
- Added a 192x192 icon and a background color to the PWA manifest so that DevDocs ticks all the requirements for an installable PWA. Not all browsers support this yet, but most major ones do. There is no explicit notification asking the user to install the PWA because there is no cross-browser way to show one, so users will have to install it manually. On Chrome Mobile and Firefox Mobile the button to install it is pretty obvious and on Chrome Desktop the user has to click on the three dots in the top-right and select "Install DevDocs". [Google is working on improving this](https://developers.google.com/web/updates/2019/06/pwa-install-addressbar). This closes #557.
- Removed all server-side styling configuration (dark/light theme, sidebar size and more) and moved it all to the client so that the amount of requests to the server is minimized. The current version of DevDocs needs to re-fetch the index page every time a styling preference is updated because some of the styling is done server-side in the `views/*.erb` files. This extends on the work done in #1011, which has already been merged into this PR.

The service worker is enabled by default in production and is opt-in in development. To enable the service worker in development, set the `ENABLE_SERVICE_WORKER` environment variable to `true` when running the server. I have updated our Dockerfiles to set this environment variable so that future versions of our Docker images will have the service worker enabled aswell.

These changes have been tested on all our supported browsers:
- Chrome 75 on Ubuntu 18.04
- Firefox 67 on Ubuntu 18.04
- Opera 62 on Ubuntu 18.04
- Chrome 75 on Android 9
- Firefox 67 on Android 9
- Safari on iOS 12.3.1
- Safari 12.1 on macOS 10.14.4
- EdgeHTML 18.18362 on Windows 10 version 1903

A live demo is running on [sw.devdocs.jmerle.dev](https://sw.devdocs.jmerle.dev/).


assets/javascripts/app/serviceworker.coffee
```diff
@@ -0,0 +1,47 @@
...
+  @isEnabled: ->
+    !!navigator.serviceWorker and app.config.service_worker_enabled
+
+  constructor: ->
+    @registration = null
+    @notifyUpdate = true
+
+    navigator.serviceWorker.register(app.config.service_worker_path, {scope: '/'})
+      .then((registration) => @updateRegistration(registration))
+      .catch((error) -> console.error 'Could not register service worker:', error)
```
j-f1: Assuming `@updateRegistration` won’t throw, this would have the same behavior. Not sure if it’s clearer or not:

```coffee
navigator.serviceWorker.register(app.config.service_worker_path, {scope: '/'})
   .then(
      (registration) => @updateRegistration(registration),
      (error) -> console.error('Could not register service worker:', error)
    )
```

assets/javascripts/app/settings.coffee
```diff
@@ -104,3 +110,38 @@ class app.Settings
...
+    @addResizerCSS()
+
+  toggleDark: (enable) ->
+    classList = document.documentElement.classList
+    classList[if enable then 'remove' else 'add']('_theme-default')
+    classList[if enable then 'add' else 'remove']('_theme-dark')
+
+  toggleLayout: (layout, enable) ->
+    classList = document.body.classList
+    classList[if enable then 'add' else 'remove'](layout) unless layout is SIDEBAR_HIDDEN_LAYOUT
```
j-f1: Since the class names on L120, L121, and L127 are written out literally, why not just do `layout is '_sidebar-hidden'` here?

assets/javascripts/app/update_checker.coffee
```diff
@@ -3,13 +3,13 @@ class app.UpdateChecker
     @lastCheck = Date.now()
 
     $.on window, 'focus', @onFocus
-    app.appCache.on 'updateready', @onUpdateReady if app.appCache
+    app.serviceWorker.on 'updateready', @onUpdateReady if app.serviceWorker
```
j-f1: The changed lines in this file could use the optional property access operator:

```coffee
app.serviceWorker?.on 'updateready', @onUpdateReady
app.serviceWorker?.update()
```

assets/javascripts/templates/error_tmpl.coffee
```diff
@@ -12,7 +12,7 @@ app.templates.notFoundPage = ->
 
 app.templates.pageLoadError = ->
   error """ The page failed to load. """,
-        """ It may be missing from the server (try reloading the app) or you could be offline.<br>
+        """ It may be missing from the server (try reloading the app) or you could be offline (try <a href="/offline">installing the documentation for offline usage</a> when online again).<br>
             If you keep seeing this, you're likely behind a proxy or firewall that blocks cross-domain requests. """,
```
j-f1:
```suggestion
             If you’re online and you keep seeing this, you’re likely behind a proxy or firewall that blocks cross-domain requests. """,
```

assets/javascripts/templates/pages/offline_tmpl.coffee
```diff
@@ -26,7 +26,7 @@ app.templates.offlinePage = (docs) -> """
   <dl>
     <dt>How does this work?
     <dd>Each page is cached as a key-value pair in <a href="https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API">IndexedDB</a> (downloaded from a single file).<br>
-        The app also uses <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache">AppCache</a> and <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API">localStorage</a> to cache the assets and index files.
+        The app also uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers">Service Workers</a> and <a href="https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API">localStorage</a> to cache the assets and index files.
```
j-f1: Just a thought: should these link to the equivalent DevDocs pages instead?

```suggestion
        The app also uses <a href="https://devdocs.io/dom/service_worker_api/using_service_workers">Service Workers</a> and <a href="https://devdocs.io/dom/web_storage_api">localStorage</a> to cache the assets and index files.
```

Also the previous line:

```html
<dd>Each page is cached as a key-value pair in <a href="https://devdocs.io/dom/indexeddb_api">IndexedDB</a> (downloaded from a single file).<br>
```

assets/javascripts/app/settings.coffee
```diff
@@ -104,3 +110,38 @@ class app.Settings
...
+
+  initLayout: ->
+    @toggleDark(@get('dark'))
+    @toggleLayout(layout, @hasLayout(layout)) for layout in @LAYOUTS
+    @addResizerCSS()
+
+  toggleDark: (enable) ->
+    classList = document.documentElement.classList
+    classList[if enable then 'remove' else 'add']('_theme-default')
+    classList[if enable then 'add' else 'remove']('_theme-dark')
```
j-f1: This can use `classList.toggle`:

```coffee
classList.toggle('_theme-default', !enable)
classList.toggle('_theme-dark', enable)
```

assets/javascripts/app/settings.coffee
```diff
@@ -104,3 +110,38 @@ class app.Settings
...
+
+  toggleLayout: (layout, enable) ->
+    classList = document.body.classList
+    classList[if enable then 'add' else 'remove'](layout) unless layout is SIDEBAR_HIDDEN_LAYOUT
+    classList[if $.overlayScrollbarsEnabled() then 'add' else 'remove']('_overlay-scrollbars')
```
j-f1: Same thing on these two lines with `toggle`:

```coffee
classList.toggle(layout, enable) unless layout is '_sidebar-hidden'
classList.toggle('_overlay-scrollbars', $.overlayScrollbarsEnabled())
```

assets/javascripts/app/settings.coffee
```diff
@@ -104,3 +110,38 @@ class app.Settings
...
+
+  addResizerCSS: ->
```
j-f1: Could this be replaced with CSS variables?

- Add the declarations below to their appropriate CSS files, reference `var(--sidebar-size)`
- Declare `--sidebar-size: 20rem` on `:root`
- Override it using `document.documentElement.style.setProperty('--sidebar-size', size + 'px')`

docs/maintainers.md
```diff
@@ -84,7 +84,7 @@ In addition to the [publicly-documented commands](https://github.com/freeCodeCam
 
 Once docs have been uploaded via `thor docs:upload` (if applicable), deploying DevDocs is as simple as running `git push heroku master`. See [Heroku's documentation](https://devcenter.heroku.com/articles/git) for more information.
 
-- If you're deploying documentation updates, verify that the documentations work properly once the deploy is done (you will need to reload [devdocs.io](https://devdocs.io/) a couple times for the application cache to update and the new version to load).
+- If you're deploying documentation updates, verify that the documentations work properly once the deploy is done (you will need to reload [devdocs.io](https://devdocs.io/) a couple times for the service worker to update and the new version to load).
```
j-f1: Might also want to recommend waiting a few seconds before reloading to allow the service worker to finish caching. After that, only one reload should be necessary.

lib/app.rb
```diff
@@ -192,35 +192,47 @@ def query_string_for_redirection
...
+                  .to_a
+                  .reject {|file| file.start_with?(App.docs_path)}
 
-    def dark_theme?
-      app_theme == 'dark'
+        return App.sprockets.pack_hexdigest(App.sprockets.files_digest(paths))
+      end
     end
 
-    def redirect_via_js(path) # courtesy of HTML5 App Cache
```
j-f1: Is this redirect still necessary with the Service Worker?

views/service-worker.js.erb
```diff
@@ -0,0 +1,54 @@
...
+});
+
+<%# Handle HTTP requests %>
+self.addEventListener('fetch', event => {
+  event.respondWith((async () => {
+    const cachedResponse = await caches.match(event.request);
+    if (cachedResponse) return cachedResponse;
+
+    try {
+      const response = await fetch(event.request);
```
j-f1: I believe this will successfully return a response with a 404 if one occurs. Would it be better to move the check on [L44](#diff-8023b663f6a9ea836432d7c5132d4c96R44) to before trying to run the `fetch`?

lib/app.rb
jmerle: Seems like it was never really necessary. All it does is ensure the correct page is loaded when you directly navigate to a page like [https://devdocs.io/offline](https://devdocs.io/offline) (instead of using the dropdown menu). The `redirect_via_js` method redirects the user to `https://devdocs.io/` and lets the JS code know which page to show through a cookie. The alternative that is used if the user has a browser that doesn't support this is to send the user to `https://devdocs.io/#/offline`, while also letting the JS code handle the routing.

It seems like the `redirect_via_js` method was added in 217d5016d05ecb15a872139336fe2453adab70d6 because the other approach seemed to cause a bug when directly navigating to certain documentation entries. I have no idea if this bug still exists, nor do I know how to test if it does because of the vague commit message, so I think it's best to keep the method in there.

assets/javascripts/app/update_checker.coffee
jmerle: Yes to the first line, no to the second line. The update() code is as follows:

```ruby
if app.serviceWorker
  app.serviceWorker.update()
else
  # check for update in another way
```

Because of the `else` clause, the optional property access operator cannot be used here.

jmerle:
Thanks for the fast yet thorough review! I've implemented all of your suggestions, lets get this deployed asap.

jmerle:
The latest commit sets the `theme-color` based on whether DevDocs is set to the dark theme or not for a better user experience on mobile devices:


I couldn't really do this in a separate PR because it relies on a method added in #1011 (which has already been merged into this PR).

cglong:
According to Microsoft's [PWABuilder](https://www.pwabuilder.com/), the demo environment received a **35/40** in the Service Worker category for failing to meet the following criterion:

>Service Worker has a `pushManager` registration

j-f1:
The Push API seems to be mainly a way to send users push notifications. That’s not something we need at the moment, but it might be useful to have an option in the future that prompts users to update their docs when updates become available.

sbrl:
It works! Thank so much :heart:

PR #1024, opened 2019-07-17
Fix OpenJDK scraper

jmerle:
This PR makes sure `read_file` in file scrapers is always passed an absolute path to the file to be read. In the current version of DevDocs this is done in the `read_file` method in [`file_scraper.rb`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/core/scrapers/file_scraper.rb) itself, but is not done in the `read_file` override in [`openjdk.rb`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/scrapers/openjdk.rb). This causes the OpenJDK scraper to ignore the first page to scrape because it can't open the file based on the relative path given to it's `read_file` method.

This PR ensures that the `read_file` method is always called with an absolute path by creating the absolute path in the `request_one` method, instead of in the `read_file` method itself.

### DIFF
--- a/lib/docs/core/scrapers/file_scraper.rb
+++ b/lib/docs/core/scrapers/file_scraper.rb
@@ -29,7 +29,7 @@ module Docs
 
     def request_one(url)
       assert_source_directory_exists
-      Response.new read_file(url_to_path(url)), URL.parse(url)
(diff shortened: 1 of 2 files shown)


j-f1:
https://travis-ci.org/freeCodeCamp/devdocs/builds/560158263#L803

PR #1046, opened 2019-08-08
cmake: add versions 3.13, 3.14, 3.15

simon04:


lib/docs/scrapers/cmake.rb
```diff
@@ -13,51 +13,66 @@ class Cmake < UrlScraper
...
     version '3.6' do
       self.release = '3.6.3'
-      self.base_url = 'https://cmake.org/cmake/help/v3.6/'
+      self.base_url = "https://cmake.org/cmake/help/v#{self.version}/"
     end
 
     version '3.5' do
       self.release = '3.5.2'
-      self.base_url = 'https://cmake.org/cmake/help/v3.5/'
+      self.base_url = "https://cmake.org/cmake/help/v#{self.version}/"
```
j-f1: Could this be extracted to the root of the class?

jmerle: Nope, if you move it to the root of the class `self.version` will not evaluate to the correct version, causing all versions to get a `base_url` of `https://cmake.org/cmake/help/v/`.

PR #1060, opened 2019-08-16
ansible: Correct 2.8 base_url and bump releases

ushuz:


jmerle:
Closing this in favor of #1061. We generally don't do patch updates unless the attribution or base url has changed.

Issue #943, opened 2018-11-25
The update all button for offline docs is not working


hjpotter92:
![image](https://user-images.githubusercontent.com/3393533/48984287-d7b45500-f11f-11e8-8c0b-b6941310eec1.png)

Even when clicking the **Update all** button, it does nothing. Manually clicking them individually works.

PR #1072, opened 2019-08-23
Fix Issue #943: Update All button not working

byted:
This change fixes https://github.com/freeCodeCamp/devdocs/issues/943

The "Update All" button uses a `<strong>` tag within the `<button>` tag, breaking the current `data-action` extraction method of looking at the target element. This PR adds an alternative way of trying to extract the action from the parent in case it can't be extracted from the target of the click.


jmerle:
Thanks for the fix!

PR #1073, opened 2019-08-28
Remove Egoist Desktop Application from Readme

Jamesking56:
It looks to be abandoned and has been archived on GitHub.


jmerle:
I'm not going to merge this because there are other projects on that list that haven't been committed to in several years. Even though egoist/devdocs-desktop has been deprecated, it still works so I don't see a reason to remove it from the list.

I did however [replace the list with a table](https://github.com/freeCodeCamp/devdocs#related-projects) in which every project includes a badge showing when its repository was last committed to, so it's more clear which projects are still being worked on and which aren't. I have also included the fact that egoist/devdocs-desktop is deprecated in its description in the table.

PR #1075, opened 2019-09-02
gitignore: add vendor from `bundle install`

simon04:


jmerle:
Thanks! I've seen this in a few earlier PR's but always removed it as I've never gotten this directory and never thought it would come from `bundle install`, but great to have it in the `.gitignore` after all.

PR #1077, opened 2019-09-02
Run `optipng -o7` on all PNG files

simon04:


jmerle:
I can't see any differences and the size of the generated spritesheet doesn't seem to change, but just to confirm, is `optipng -o7` lossless or not?

j-f1:
> OptiPNG is a PNG optimizer that recompresses image files to a smaller size, without losing any information.

Maybe this should be run programmatically on the generated sprite sheet instead?

jmerle:
That quote answers my question, thanks.

As for automatically running it on the generated sprite sheet: we already use [image_optim](https://github.com/toy/image_optim) in one of the core filters which extracts images out of scraped html. I think we can reuse some of that code, along with it's image_optim configuration, in the `sprites:generate` Thor task. I'll see if that makes a different in sprite sheet size in a few minutes.

jmerle:
OptiPNG is now being ran in production on the generated spritesheets. Here's the change in size:

```
docs.png: 81,0 KiB -> 69,8 KiB
docs@2x.png: 213,4 KiB -> 183,9 KiB
```

It's disabled in development and when running tests because it's quite slow.

PR #1080, opened 2019-09-05
Update PostgreSQL to version 11 (release 11.5)

filalex77:

### DIFF
--- a/assets/javascripts/templates/pages/about_tmpl.coffee
+++ b/assets/javascripts/templates/pages/about_tmpl.coffee
@@ -593,7 +593,7 @@ credits = [
     'https://raw.githubusercontent.com/ponylang/ponyc/master/LICENSE'
   ], [
     'PostgreSQL',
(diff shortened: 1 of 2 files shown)


filalex77:
I'm having trouble building and testing this PR locally. Here's the error I have when running `bundle exec thor docs:generate postgresql --version 11`: https://gist.github.com/filalex77/f8ca779ee47f488127ca71ec59c63e5d.

Would be great if someody could help me on this :)

jmerle:
Be careful when copying and pasting existing code. In this case, the `base_url` for version 11 should be `https://www.postgresql.org/docs/11/` instead of `https://www.postgresql.org/docs/11/static/` (try going to the latter in the browser, it'll redirect to the former).

filalex77:
@jmerle Oh, I noticed that, but thought that was the expected behavior (`https://www.postgresql.org/docs/10/static/` did the same for me)

I'll update that.

jmerle:
Thanks, looks good! The PostgreSQL 11.5 docs are now available at [https://devdocs.io/postgresql~11/](https://devdocs.io/postgresql~11/).

PR #1086, opened 2019-09-11
Add support for prefers-color-scheme media query

### LINKED ISSUE #1012: Add support for prefers-color-scheme media query
It would be great to add support for `prefers-color-scheme` to automatically match a user's active colour scheme.

My personal reason for wanting this (rather than manually setting the colour scheme) is that I have dark mode toggled automatically at susnset on my Mac using Flux, and it would be great if DevDocs followed it rather than having to always be either light or dark.

I would suggest replacing the "Enable dark theme" checkbox with a dropdown  with three options:
- Automatic (default - this option  would respect `prefers-color-scheme`)
- Dark
- Light

To enable these manual overrides to work, the  implementation would presumably use `window.matchMedia` and then toggle the colour scheme in JS as is currently done, rather than using media queries in CSS.

simon04:
Fixes #1012.

### DIFF
--- a/assets/javascripts/app/settings.coffee
+++ b/assets/javascripts/app/settings.coffee
@@ -32,10 +32,16 @@ class app.Settings
     manualUpdate: false
     schema: 1
     analyticsConsent: false
+    theme: 'auto'
 
   constructor: ->
     @store = new CookiesStore
     @cache = {}
+    @autoSupported = window.matchMedia('(prefers-color-scheme)').media != 'not all'
+    if @autoSupported
+      @darkModeQuery = window.matchMedia('(prefers-color-scheme: dark)')
+      @darkModeQuery.onchange = -> @setTheme(@get('theme'))
+
 
   get: (key) ->
     return @cache[key] if @cache.hasOwnProperty(key)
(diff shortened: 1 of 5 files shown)


jmerle:
I haven't checked it out locally yet, but I see a few issues:
* Not all browsers support `prefers-color-scheme`, including Edge which is one of the browsers DevDocs is supported on. We should hide the "Automatic" option and default to the light theme like we do now on browsers which don't support it. [Seems to be do-able in JavaScript](https://web.dev/prefers-color-scheme#finding-out-if-dark-mode-is-supported-by-the-browser).
* `prefers-color-scheme` can change without having to reload, yet from looking at it, the code in this PR requires a refresh for changes in `prefers-color-scheme` to be picked up. Since we already use CSS variables, it should be do-able to implement part of this PR in CSS (like, use `prefers-color-scheme` in CSS only if body has class `_theme-automatic` or something alike) so that changes to the user's preferred color scheme are picked up without requiring a refresh.

What are your thoughts on this?

j-f1:
> like, use `prefers-color-scheme` in CSS only if body has class `_theme-automatic` or something alike

```css
@media (prefers-color-scheme: dark) {
  html._theme-auto {
    /* ... dark vars */
  }
}
```

j-f1:
@simon04 Can you check the updated version and see if it works for you?

simon04:
@j-f1, I've tested your changes successfully using `chromium --force-dark-mode` (version 79.0.3945.117) as well as Firefox 72.0.1 using https://bugzilla.mozilla.org/show_bug.cgi?id=1519547#c7

j-f1:
Thanks for adding this!

PR #1090, opened 2019-09-24
Update Crystal documentation (0.31.0)

filalex77:

---

I made the decision to create a new version because of [breaking changes in Crystal 0.31](https://github.com/crystal-lang/crystal/releases/tag/0.31.0) (like removing the `Markdown` module, for example).


jmerle:
Updated and deployed. In the future, please don't tick the "Tested the changes locally to ensure" checkbox when you haven't actually done so or when one of the items underneath it isn't true.

PR #1091, opened 2019-10-01
Update Phoenix documentation (1.4.10)

Br1ght0ne:

---
This change is [Reviewable](https://reviewable.io/reviews/freecodecamp/devdocs/1091)

### DIFF
--- a/lib/docs/scrapers/phoenix.rb
+++ b/lib/docs/scrapers/phoenix.rb
@@ -1,7 +1,7 @@
 module Docs
   class Phoenix < UrlScraper
     self.type = 'elixir'
-    self.release = '1.3.4'
+    self.release = '1.4.10'
     self.base_url = 'https://hexdocs.pm/'
     self.root_path = 'phoenix/Phoenix.html'
     self.initial_paths = %w(


lib/docs/scrapers/phoenix.rb
```diff
@@ -1,9 +1,8 @@
 module Docs
   class Phoenix < UrlScraper
     self.type = 'elixir'
-    self.release = '1.3.4'
     self.base_url = 'https://hexdocs.pm/'
-    self.root_path = 'phoenix/Phoenix.html'
+    self.root_path = "phoenix/#{release}/Phoenix.html"
```
jmerle: This isn't going to work because self.root_path is evaluated before you create the different releases, which means the root path will always be "phoenix//Phoenix.html".

lib/docs/scrapers/phoenix.rb
```diff
@@ -47,6 +46,14 @@ class Phoenix < UrlScraper
       end
     }
 
+    version '1.3' do
+      self.release = '1.3.4'
+    end
+
+    version '1.4' do
+      self.release = '1.4.10'
+    end
```
jmerle: Why do we need two separate releases for a minor update? I don't see any major breaking changes between 1.3.4 and 1.4.10, so I think keeping it to one version is a better idea.

filalex77: Sorry, I think I misread the release notes. There is indeed no breaking changes to the docs.

filalex77:
Thanks for the review. I'm still rusty at the core principles here, so it helps a lot.

filalex77:
@jmerle I'm pretty sure I didn't have this error. I'll do a rebuild and check the filter too.

scherepn:
This PR was superseded by PR #1262 and can be closed.

PR #1099, opened 2019-10-05
go: update to latest major version

owen5mith:


jmerle:
Thanks! Merged and deployed.

Issue #1101, opened 2019-10-05
Option to auto-hide the side bar doesn't work after reloading


gsz:
There's a setting in the preferences to automatically show and hide the sidebar, which also causes the doc display area to fill the entire screen. I can activate it and at first it works like it should. However, after I reload the page, it goes back to default behavior with sidebar always being displayed. The option is still checked in the preferences. Visiting and immediately leaving the preferences screen causes it to start working again.

The setting is persistent, so it should be respected immediately after loading the page.

I observed this on Firefox 69.0 and Chromium 76.0.3809.100.

byted:
I looked a little bit into it and it seems to be caused by https://github.com/byted/devdocs/blob/bdb5c40b13c7a14d8e5d86a779044bcfd379b5e4/assets/javascripts/app/settings.coffee#L132 
`toggleLayout` is used during the initial load of the app to read the stored layout(s) and apply them - however, it specifically filters out auto-hide. As far as I can tell that's done b/c this method is also used to apply layouts when navigating to the settings page in which case the auto-hide layout should be disabled as the sidebar is always shown on the settings page.

If I have some spare time this week I'll fix it.

StevenJokes:
The hiding isn't automatic when I used. But thanks to @gsz ,I have made it in perferences

PR #1102, opened 2019-10-08
Fix issue #1101: reloading breaks auto-hide layout

byted:
This change fixes https://github.com/freeCodeCamp/devdocs/issues/1101 by setting the `_auto-hide` layout based on current location instead of simply by name. My comment to the tickets gives a little more background.


jmerle:
Thanks for the quick fix!

PR #1109, opened 2019-10-13
Update Rails documentation (6.0.0)

TheWillGabriel:

As mentioned in https://github.com/freeCodeCamp/devdocs/issues/1108, I've added some files to the skip options for the scraper that are erroneously added from internal URLs. I've confirmed that both of the actual files referenced in those anchor links (active_record_migrations and association_basics) are properly scraped.


jmerle:
Thanks! Merged and deployed to [https://devdocs.io/rails~6.0/](https://devdocs.io/rails~6.0/).

PR #1107, opened 2019-10-12
Add Sequelize documentation

waldyrious:
Add scraper for [Sequelize](https://sequelize.org/), whose docs live [here](https://sequelize.org/master/index.html). Thanks @jmerle for the precious help on the [Gitter chat room](https://gitter.im/FreeCodeCamp/DevDocs), and @hugopeixoto for the offline assistance.

Following the PR checklist, I confirm that I have:

- [x] Tested the scraper on a local copy of DevDocs
- [x] Ensured that the docs are styled similarly to other docs on DevDocs
- [x] Added these files to the <code>public/icons/sequelize/</code> directory:
  - [x] `16.png`: a 16×16 pixel icon for the doc
  - [x] `16@2x.png`: a 32×32 pixel icon for the doc
  - [x] `SOURCE`: A text file containing the URL to the page the image can be found on or the URL of the original image itself

Screenshot_2019-10-13 Sequelize documentation — DevDocs

There are a couple of unsolved issues still:
- The page titles are not all formatted in a consistent way, but I'm not sure what's the best way to harmonize them.
- I couldn't figure out how to group pages within the same category (e.g. the various errors within the API Reference, as seen in the "Directories" box at the top right side of https://sequelize.org/master/identifiers.html), which makes the navigation in DevDocs less helpful than the official documentation's sidebar

Any help in sorting these out would be appreciated.

PR #1117, opened 2019-10-26
Remove unused code in Sequelize's clean_html filter

waldyrious:
Also fix it's --> its in code comments.

This is a follow-up to #1107.


waldyrious:
For future reference: this code was not used because the pages it applied to are not scraped (they were in the initial proposed version of the scraper). Quoting from #1107:

> Removed the source files from the generated docs, we generally don't include those in DevDocs.

[Here's the diff](https://github.com/freeCodeCamp/devdocs/pull/1107/commits/089aa6158aa9997da4f56eb40d2b81b6b91bf8f9#diff-95a76bcefa69288759fe138eb2210e3fL13-L14) of the change quoted above.

PR #1118, opened 2019-10-29
Update Angular documentation to include 7 and 8

kaltepeter:

First time updating a doc. I updated angular to the latest which was two major versions behind. Adds angular 7 & 8.


j-f1:
Thank you for contributing to DevDocs! I made a couple of changes in https://github.com/freeCodeCamp/devdocs/commit/84b83050346265a3e011ebe21e93f7c5a4496bf9 and should have the new docs deployed shortly.

PR #1119, opened 2019-10-29
Update mongoose version to 5.7.7

kaltepeter:

This seems very straight forward. Had to switch docs url to https instead of http.


j-f1:
Thank you for contributing! https://devdocs.io/mongoose/

PR #1143, opened 2019-12-17
Fix "doc is undefined" error for local install

waffshappen:
Having installed this locally
(Fedora 31, inside a vm, apache proxypass / -> 9292/)
i encountered an error that the variable "doc" was undefined (null), likely because coffeescript extends this to walking over the array and some element being undefined? Either way, checking if doc is not null (and thus if ref[i] isnt producing a null result) makes the page work again.

### DIFF
--- a/assets/javascripts/app/app.coffee
+++ b/assets/javascripts/app/app.coffee
@@ -92,7 +92,8 @@
   bootAll: ->
     docs = @settings.getDocs()
     for doc in @DOCS
-      (if docs.indexOf(doc.slug) >= 0 then @docs else @disabledDocs).add(doc)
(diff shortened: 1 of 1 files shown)


waffshappen:
>   File: bundler-2.1.0.gem
> 
> bundler's executable "bundle" conflicts with /home/travis/.rvm/rubies/ruby-2.6.5/bin/bundle
> 
> Overwrite the executable? [yN]  

Travis seems to also need a fix

bigkatspence:
@waffshappen  did you check to make sure your nods.js.react config was set in the right patterns and do you have a screenshot of your code with the errors present?

PR #1155, opened 2020-01-10
Allow styles in <dt> elements

### LINKED ISSUE #1154: Fix header formatting in Sphinx-based docs
The Sphinx `clean_html` filter removes all formatting from `<dt>` elements:

https://github.com/freeCodeCamp/devdocs/blob/58a0120697a701f2cb9ca87061ec7ef6fde3b013/lib/docs/filters/sphinx/clean_html.rb#L64

However, this is not entirely desirable since the formatting makes things look better. For example, the [header for `BaseException`](http://devdocs.localhost/python~3.8/library/exceptions#BaseException) ([original](https://docs.python.org/3.8/library/exceptions.html#BaseException)) is transformed from this:

> *exception* **`BaseException`**

to this:
> `exception BaseException`

cc @Thibaut since you might have a reason for implementing this this way. (no worries if you don’t reply though)

j-f1:
I have not yet checked to see if this breaks things. Fixes #1154.

Steps
- check if doc includes `<dt>`
- check each `<dt>` to see if it has unwanted content

Potentially affected docs:
- [ ] ansible
- [ ] bottle
- [ ] cmake
- [ ] codeigniter
- [ ] django
- [ ] falcon
- [ ] godot
- [ ] julia
- [ ] matplotlib
- [ ] numpy
- [ ] pandas
- [ ] python
- [ ] scikit_image
- [ ] scikit_learn
- [ ] statsmodels

### DIFF
--- a/lib/docs/filters/sphinx/clean_html.rb
+++ b/lib/docs/filters/sphinx/clean_html.rb
@@ -57,14 +57,6 @@ module Docs
           node.parent.before(node.content).remove
         end
 
-        css('dt').each do |node|
-          next unless node['id'] || node.at_css('code, .classifier')
-          links = []
-          links << node.children.last.remove while node.children.last.try(:name) == 'a'
-          node.inner_html = "<code>#{CGI::escapeHTML(node.content.strip)}</code> "
-          links.reverse_each { |link| node << link }
-        end
-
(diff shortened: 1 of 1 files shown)


arun0930:
Yyy

PR #1144, opened 2019-12-18
Upgrade TensorFlow 2.1

lgeiger:
This PR upgrades TensorFlow documentation to version 2.1.

I also removed version 1.8 since it is not available on the website anymore. After this PR one could easily add support for version 2.0, 1.15 - 1.10.

Closes #1141
Closes #1010

## PR Checklist

- [x] Updated the versions and releases in the scraper file
- [x] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches it's data in `self.attribution`
- [x] Ensured the icons and the `SOURCE` file in <code>public/icons/*your_scraper_name*/</code> are up-to-date if the documentation has a custom icon
- [x] Ensured `self.links` contains up-to-date urls if `self.links` is defined
- [x] Tested the changes locally to ensure:
  - The scraper still works without errors
  - The scraped documentation still looks consistent with the rest of DevDocs
  - The categorization of entries is still good

PR #1165, opened 2020-01-25
Split Tensorflow into separate Python and C++ docs

lgeiger:
@j-f1 Thanks for merging #1144

This PR splits the TensorFlow docs into separate Python, C++ and Guides pages which allows to include docs for TensorFlow versions 2.1, 2.0 and 1.15 so we can start to properly version the TensorFlow docs.

## Checklist


j-f1:
It looks like the C++ docs dump everything into a single category. Could you make it more like the Python version, which splits up the entries by category?

j-f1:
Also, it looks like some of the pages in the guide are duplicated. Is it even worth including the guide, since DevDocs doesn’t support full text search.

lgeiger:
Sorry for the long wait.

> Finally, can you add '.tfo-notebook-buttons img', '.tfo-notebook-buttons>:first-child' to the list of things to remove in clean_html.rb line 7? This improves the display of the action links, especially in the guide.

👍 I removed the icons and the broken link in 7c516024f1dbd78f443e466bc7fdc3f8b63cdbcd

> Is it even worth including the guide, since DevDocs doesn’t support full text search.

I'd be very happy to remove the guides, I only included them because they were already present.  I don't think they add much value and it is not really possible to version them correctly either.

> Could you make it more like the Python version, which splits up the entries by category?

I tried to properly structure the C++ docs, but unfortunately the structure cannot be inferred by the slug, and inferring the structure from the navigation sidebar like it was done [before](https://github.com/freeCodeCamp/devdocs/pull/1144/files#r359131678) doesn't work for me. I haven't worked with Ruby before, so I am not sure how this can be fixed. Since this problem also exists on master I think this should be solved in a future PR.

j-f1:
I unfortunately won’t be able to merge this in the near future because I don’t have great Internet at my house so it can’t really scrape these giant docs but I’ll keep this in mind for whenever I’m somewhere with good internet.

lgeiger:
> I unfortunately won’t be able to merge this in the near future because I don’t have great Internet at my house so it can’t really scrape these giant docs but I’ll keep this in mind for whenever I’m somewhere with good internet.

No problem, thanks for the heads up. Let me know if there is anything I can help with.

Should I remove the guides as part of this PR? This should also make scraping easier.

j-f1:
Sure! I still had trouble scraping the Python and C++ docs though.

lgeiger:
:+1: I removed the guides in 0e3cdfd25e5f6b91d7c703b331851113d348c349

simon04 (changes requested):
The scraper does not run.
- ```SyntaxError: /home/simon/src/devdocs/lib/docs/filters/tensorflow/entries.rb:18: syntax error, unexpected `elsif'
        elsif slug.start_with?('tf/audio')``` (easy to fix)
- `index.html` does no longer exist → `self.root_path = '/tf'`
- ```  https://www.tensorflow.org/versions/r2.1/api_docs/python/tf
  HTML::Pipeline::Filter::InvalidDocumentException: HTML::Pipeline::Filter::InvalidDocumentException
  /home/simon/src/devdocs/lib/docs/core/filter.rb:6:in `css'
  /home/simon/src/devdocs/lib/docs/filters/tensorflow/clean_html.rb:7:in `call'
```

I'm happy to review and update the docs once we have a working scraper. In the meanwhile, 2.3 has been released.

**Comment by lgeiger** [comment]
@simon04 Thanks for taking a look. I updated the PR to reflect the latest changes on the TF docs website. I wasn't able to check the generated docs yet, but feel free to takeover this PR. I might be able to take a proper look at it next weekend.

File: lib/docs/scrapers/tensorflow/tensorflow.rb [hunk 1]
```diff
@@ -14,24 +14,14 @@ class Tensorflow < UrlScraper
...
-
-    version '2.0' do
-      self.release = '2.0.0'
-      self.base_url = "https://www.tensorflow.org/versions/r#{version}/api_docs/python"
-    end
-
-    version '1.15' do
-      self.release = '1.15.0'
-      self.base_url = "https://www.tensorflow.org/versions/r#{version}/api_docs/python"
+    for version in ['2.3', '2.2', '2.1', '2.0', '1.15'] do
```
simon04: The statement `version '2.3' do … end` invokes the method `version` from `lib/docs/core/doc.rb` and cannot be replaced by a loop like this. I've fixed this syntax

simon04 (approved):
Thank you!

PR #1166, opened 2020-01-28
Add Django alias 'dj'

adamchainz:
Django is used by about 45% of all Python developers ( https://www.jetbrains.com/research/python-developers-survey-2018/ ) so an alias seems appropriate :)


j-f1:
This shouldn’t be necessary — if you use the `dj` shortcut it should already work.

PR #1169, opened 2020-01-29
Update Godot versions

ignaloidas:

### DIFF
--- a/lib/docs/scrapers/godot.rb
+++ b/lib/docs/scrapers/godot.rb
@@ -17,12 +17,12 @@ module Docs
     options[:attribution] = ->(filter) do
       if filter.subpath.start_with?('classes')
          <<-HTML
-          &copy; 2014&ndash;2019 Juan Linietsky, Ariel Manzur, Godot Engine contributors<br>
+          &copy; 2014&ndash;2020 Juan Linietsky, Ariel Manzur, Godot Engine contributors<br>
           Licensed under the MIT License.
         HTML
       else
         <<-HTML
-          &copy; 2014&ndash;2019 Juan Linietsky, Ariel Manzur and the Godot community<br>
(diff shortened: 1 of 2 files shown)


ignaloidas:
Unable to test the changes locally currently, but looks like everything should be fine.

j-f1:
It looks like the 3.0 series is no longer available from https://godotengine.org/download. Should we still host the docs?

ignaloidas:
After some discussion in Godot's IRC 3.0 and 3.1 versions are dropped as 3.0 is not really used anymore and 3.1 is compatible with 3.2 docs

allanjoh:
I was not able to get this running locally: 
```
 bundle exec thor docs:generate godot --version=3.2
/!\ WARNING /!\

Some scrapers send thousands of HTTP requests in a short period of time,
which can slow down the source site and trouble its maintainers.

Please scrape responsibly. Don't do it unless you're modifying the code.

To download the latest tested version of this documentation, run:
  thor docs:download godot

Proceed? (y/n) y
[###################################################################################################] [1/1] [100.00%] [00:00] [00:00] [1.26/s]Failed! (try running with --debug for more information)
```

However, if we change the protocol to `https` then it succeeds

lib/docs/scrapers/godot.rb
```diff
@@ -28,18 +28,13 @@ class Godot < UrlScraper
...
 
-    version '3.1' do
-      self.release = '3.1.1'
-      self.base_url = "http://docs.godotengine.org/en/#{self.version}/"
-    end
-
-    version '3.0' do
-      self.release = '3.0.6'
+    version '3.2' do
+      self.release = '3.2.1'
```
Calinou, line 32: Godot 3.2.3 is the latest stable version now:

```suggestion
      self.release = '3.2.3'
```

simon04 (approved):

PR #1216, opened 2020-04-21
update php documentation (7.4.5)

jhdxr:

### DIFF
--- a/lib/docs/filters/php/fix_urls.rb
+++ b/lib/docs/filters/php/fix_urls.rb
@@ -3,7 +3,7 @@ module Docs
     class FixUrlsFilter < Filter
       def call
         html.gsub! File.join(Php.base_url, Php.root_path), Php.base_url
-        html.gsub! %r{https://secure\.php\.net/manual/en/([^"']+?)\.html}, 'https://secure.php.net/manual/en/\1.php'
+        html.gsub! %r{https://www\.php\.net/manual/en/([^"']+?)\.html}, 'https://www.php.net/manual/en/\1.php'
         html
       end
     end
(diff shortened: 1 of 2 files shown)


rickmacgillis:
I was just about to write an update for the PHP versions, but it seems that @jhdxr beat me to it, so that saves me some work.

Any chance someone can inspect his PR and merge it for us?

jhdxr:
Hi @rickmacgillis 
> I just noticed that he removed the old version of the docs instead of adding a versions list. I'm going to have a go at it to make it so people can select which version(s) they want to install.
The reason I do so is it's how devdocs deal with php document now, and so does php.net. 

Since php document is basicly a reference for all the php functions. It indicates since when functions are avaliable or removed rather than providing documentation for different php versions.

zicjin:
@jhdxr May I ask this PR support from https://www.php.net/distributions /manual/php_manual_en.tar.gz Download the offline documentation? I download the offline document and execute "thor docs:generate php", why do I get a lot of code that doesn't match the actual html structure? clear_html.rb:16 at_css('.manualnavbar') Object Non-existent.

simon04 (changes requested):
I cannot build the docs after downloading the 2020-11-13 version from php.net via `mkdir -p docs/php/; curl https://www.php.net/distributions/manual/php_manual_en.tar.gz | tar xz --strip-components=1 --directory=docs/php/`

jhdxr:
Sorry I'm going to abandon this PR. It seems the php manual updated its css and dom, so besides the contents changes (including new pages, excluding removed pages, like what this PR does), the more important thing is to update the filters to make it compatibile with the latest manual. I don't have enough time to check and implement those changes currently. Maybe @rickmacgillis  can help on this? If you are still interested in it.

simon04:
> If you are still interested in it.

If the question was directed to the @freeCodeCamp/devdocs team: yes, definitively!

jhdxr:
@simon04 Really glad to hear that. Hope to see it get updated soon. Thanks.

PR #1217, opened 2020-04-21
Update pandas documentation (update 0.25.0 to 0.25.3, add 1.0.3)

jbfiot:


jbfiot:
This seems to require a bit more work than expected. 

The format of the webpage has changed between [0.25.3](https://pandas.pydata.org/pandas-docs/version/0.25.3/) and [1.0.3](https://pandas.pydata.org/pandas-docs/version/1.0.3/), so the filters and/or scraper would need to be adapted.

PR #1089, opened 2019-09-24
Update Nim documentation (1.0.0)

Br1ght0ne:
- [X] Updated the versions and releases in the scraper file
- [X] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches it's data in `self.attribution`
- [X] Ensured the icons and the `SOURCE` file in <code>public/icons/*your_scraper_name*/</code> are up-to-date if the documentation has a custom icon
- [X] Ensured `self.links` contains up-to-date urls if `self.links` is defined
- [ ] Tested the changes locally to ensure:
  - The scraper still works without errors
  - The scraped documentation still looks consistent with the rest of DevDocs
  - The categorization of entries is still good

---

Currently emits an error on a single page:

```
ERROR:
  https://nim-lang.org/docs/strutils.html
  ArgumentError: invalid byte sequence in UTF-8
```

Everything else is good.

---
This change is [Reviewable](https://reviewable.io/reviews/freecodecamp/devdocs/1089)

PR #1218, opened 2020-04-23
Update Nim docs to v1.2.0

kaushalmodi:
I am unable to build devdocs locally. I have only updated the version from 0.19.0 to 1.2.0 (it was long overdue).

I do not know how to ensure the below.

- [ ] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches it's data in `self.attribution`


genotrance:
Any update?

simon04:
Solely updating `self.release` does not suffice in this case.

Duplicate of #1089

PR #1228, opened 2020-05-18
Added new laravel versions and fixed scraper

rickmacgillis:

I've updated the Laravel scraper to handle the new website format on their site, and to add the following versions. The scraper is still backwards compatible with older versions, as they all use the same site layout.

* Laravel 5.8
* Laravel 6.0
* Laravel 7.0

### DIFF
--- a/lib/docs/filters/laravel/clean_html.rb
+++ b/lib/docs/filters/laravel/clean_html.rb
@@ -56,7 +56,7 @@ module Docs
       end
 
       def other
-        @doc = at_css('article')
+        @doc = at_css('.docs_body')
 
         # Clean up headings
         css('h2 > a').each do |node|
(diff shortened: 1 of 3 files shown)


Rai:
Would be great to have this...

lib/docs/scrapers/php.rb
```diff
@@ -20,11 +24,6 @@ class Php < FileScraper
       reserved.interfaces.html
       reserved.variables.html)
 
-    self.links = {
-      home: 'https://secure.php.net/',
-      code: 'https://git.php.net/?p=php-src.git;a=summary'
-    }
-
```
simon04, line 27: The changes to php.rb are unrelated. I'm going to cut them off this PR…

simon04 (approved):
Awesome, thank you! I've updated the release numbers and added version 8.

PR #1270, opened 2020-09-09
Add Elisp docs for version 26.3

MasterEnoc:
Add Elisp docs


simon04:
I don't know anything about elisp and how its documentation is used.

As an outsider, I find the 928 alphabetical entries in the Elisp / Manual intimidating. Would it make sense to group it (using the headings from the manual index)? – 


The navigation links should probably be removed –

MasterEnoc:
PR updates. This pr was initially for elisp 26.3, now elisp is on 27.1 version and the web page received some updates.

This scraper entries were inspired by the bash entries, both share similar approach since both are manual and api documentation.

Elisp has a lot of documentation due emacs. It is the main programming language of emacs and shares some emacs' documentation plus low level fuction that emacs' manual does not have.

Please review the devdocs entry and comment if it would be better to group as you commented.

simon04 (approved):
Thank you!

PR #1286, opened 2020-10-22
Update the Go version to 1.15

ocavue:

### DIFF
--- a/lib/docs/scrapers/go.rb
+++ b/lib/docs/scrapers/go.rb
@@ -1,7 +1,7 @@
 module Docs
   class Go < UrlScraper
     self.type = 'go'
-    self.release = '1.13'
+    self.release = '1.15'
     self.base_url = 'https://golang.org/pkg/'
     self.links = {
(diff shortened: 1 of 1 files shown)


ocavue:
There are two things I'd like to confirm:

1. The document and `Gemfile` require Roby v2.6. But since I didn't have Ruby on my computer, I used Github Codespace to develop this PR, which has only Rudy v2.7 installed. `bundle exec rackup --host 0.0.0.0` works fine so I think I can ignore the version conflict?
image
2. After edit `scrapers/go.rb` and running `thor docs:generate`, the document of Go v1.15 has been generated successfully. However, the generated files (`public/docs/go`) are ignored by Git and the [document](https://github.com/freeCodeCamp/devdocs/blob/master/.github/CONTRIBUTING.md#updating-existing-documentations) didn't say anything about adding generated documents. Thus, I guess I don't need to include those generated file into the Git repository?

simon04:
Thanks! The docs are now live on https://devdocs.io/go/

PR #1287, opened 2020-10-30
Update lua.rb

stuartpb:
This adds Lua 5.4, bumps Lua 5.3 to (probably) the last version, and updates the copyright.


jmerle:
Thanks! The docs for Lua 5.4 are now live on https://devdocs.io/lua~5.4/.

PR #1293, opened 2020-11-06
/about: Obtain credits from docs metadata

simon04:
Request for comments.

We could store the attribution/license information in the docs metadata. For leaflet, this would be
```diff
 {
   "name": "Leaflet", "slug": "leaflet~1.6", "type": "simple",
   "links": {
     "home": "https://leafletjs.com/",
+    "license": "https://raw.githubusercontent.com/Leaflet/Leaflet/master/LICENSE",
     "code": "https://github.com/Leaflet/Leaflet"
   },
   "version": "1.6", "release": "1.6.0",
+  "attribution": "      &copy; 2010&ndash;2019 Vladimir Agafonkin<br>\n      &copy; 2010&ndash;2011, CloudMade<br>\n      Maps &copy; OpenStreetMap contributors.\n",
+  "license": "BSD",
   "mtime": 1604697350, "db_size": 722507
 }
```

This allows us to get rid of ½ step from the [updating checklist](https://github.com/freeCodeCamp/devdocs/blob/master/.github/CONTRIBUTING.md#updating-existing-documentations). However, the main https://devdocs.io/docs.json file would get (considerably?) larger.


simon04:
I've simplified the 2020 concept: The license is contained in doc's `:attribution` string anyway, and we can link to the source code repository. So we do not need to maintain any additional information.

localhost_9292_

PR #1294, opened 2020-11-06
Fix broken get_latest_version implementations

jmerle:
This PR fixes the broken `get_latest_version` implementations. This includes all implementations which throw an error and the Dart, PHP and Godot implementations which return the wrong output.


simon04:
Three scraper's get_latest_version got updated via my ancient PR https://github.com/freeCodeCamp/devdocs/pull/1177.

simon04 (approved):
Awesome, thank you!

PR #1296, opened 2020-11-09
Fix get_latest_version procedure that were showing bad information

MasterEnoc:
- markdown: update scraper and fix get_latest_version

- rails: fix get_latest_version, it was showing an outdated
  version as the latest version

- apache: fix get_latest_version, it was showing and
  outdated version as the latest version

- cakephp: fix get_latest_version, it was showing and
  outdated version as the latest version

### DIFF
--- a/lib/docs/scrapers/markdown.rb
+++ b/lib/docs/scrapers/markdown.rb
@@ -2,7 +2,6 @@ module Docs
   class Markdown < UrlScraper
     self.name = 'Markdown'
     self.type = 'simple'
-    self.release = '1.0.1'
     self.base_url = 'https://daringfireball.net/projects/markdown/syntax'
 
     html_filters.push 'markdown/clean_html', 'markdown/entries'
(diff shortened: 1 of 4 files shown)


lib/docs/scrapers/markdown.rb
```diff
@@ -2,7 +2,8 @@ module Docs
   class Markdown < UrlScraper
     self.name = 'Markdown'
     self.type = 'simple'
-    self.base_url = 'http://daringfireball.net/projects/markdown/syntax'
+    self.release = '1.0.1'
+    self.base_url = 'https://daringfireball.net/projects/markdown/syntax'
```
simon04, line 6: I don't think the Markdown specs are versioned. AFAIK, version 1.0.1 refers to the Perl reference implementation…

simon04 (changes requested):
Can you please undo the changes to markdown (see my comment)? The other changes are great.

MasterEnoc:
should I undo all the changes of markdown? I can leave `self.release = '1.0.0'` to avoid showing it in outdated documentation

simon04:
I'd not invent a versioning number when the specification does not provide any. Just my 2¢…

MasterEnoc:
Fixed, please proceed to review it

simon04 (approved):

PR #1309, opened 2020-11-13
openjdk: merge 8, update release 8u272

simon04:

Fixes https://github.com/freeCodeCamp/devdocs/issues/967.


simon04:
How are devdocs and the deployment affected when merging `version '8'` `version '8 GUI'` `version '8 Web'` into `version '8'`?

scherepn:
The package sets were split up due to cache size concerns at the time. The [original PR](https://github.com/freeCodeCamp/devdocs/pull/602) has the full discussion about it.

PR #1315, opened 2020-11-14
Fix Codeigniter scraper

scherepn:

There are other problems with the documentation, but that's because the source documentation is malformed.

### DIFF
--- a/assets/stylesheets/pages/_sphinx.scss
+++ b/assets/stylesheets/pages/_sphinx.scss
@@ -4,6 +4,7 @@
   h4 { font-size: 1em; }
   > dl:not(.docutils) > dt { @extend %block-label, %label-blue; }
   dd > dl:not(.docutils) > dt { @extend %block-label; }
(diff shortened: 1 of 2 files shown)


simon04:
Please link to a page that is malformed now. I clicked through a few pages and couldn't spot any (obvious) mistake.

scherepn:
It's not the appearance of the HTML pages that are malformed (they all look perfectly fine). It's the entity extractor that has problems. Look under the `Email` category to see how this update cleans it up.

There's also one missing page: https://devdocs.io/codeigniter~4/outgoing/view_renderer
The source page https://codeigniter.com/userguide4/outgoing/view_renderer.html exists, and the reason it's not present in the generated documentation set is because the entity extractor hit a runtime error. This isn't due to a code issue, but rather the source file missing properly generated id tags.

simon04 (approved):
Thank you for the detailed explanation!

PR #1325, opened 2020-11-16
Update Gcc scraper

MasterEnoc:
- Add 10.2.0 version

- Update previous versions

### DIFF
--- a/lib/docs/scrapers/gnu/gcc.rb
+++ b/lib/docs/scrapers/gnu/gcc.rb
@@ -69,66 +69,66 @@ module Docs
     end
 
     version '8' do
-      self.release = '8.4.0'
+      self.release = '8.3.0'
       self.base_url = "https://gcc.gnu.org/onlinedocs/gcc-#{release}/gcc/"
     end
 
     version '8 CPP' do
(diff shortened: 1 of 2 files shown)


lib/docs/scrapers/gnu/gcc.rb
```diff
@@ -46,79 +46,89 @@ class Gcc < Gnu
...
-      self.release = '9.2.0'
+      self.release = '9.3.0'
       self.base_url = "https://gcc.gnu.org/onlinedocs/gcc-#{release}/cpp/"
 
       options[:replace_paths] = CPP_PATHS
     end
 
     version '8' do
-      self.release = '8.3.0'
+      self.release = '8.4.0'
```
simon04, line 72: Updating all previous releases requires (me) to download and re-generate all docs (for a consistent state). Unless there are substantial changes to older versions, please refrain from updating those versions (maybe except for the second to newest one). Thank you!

simon04 (changes requested):

MasterEnoc:
Fixed, proceed to review it

simon04 (approved):
Thank you!

PR #1332, opened 2020-11-17
Fix TensorFlow Python frontpage

lgeiger:
This PR fixes the TensorFlow Python index page which is currently broken: https://devdocs.io/tensorflow~2.3/

/cc @simon04


simon04 (approved):
Thank you!

PR #1337, opened 2020-11-18
feat: Docker development environment

ojeytonwilliams:
This should make it a little easier to get started as a new contributor, as well as cut out 'works on my machine' bugs.

@freeCodeCamp/devdocs let me know what you think.  Does this all work as you expect?


.github/CONTRIBUTING.md
```diff
@@ -12,7 +12,8 @@ Want to contribute? Great. Please review the following guidelines carefully and
 6. [Updating existing documentations](#updating-existing-documentations)
 7. [Other contributions](#other-contributions)
 8. [Coding conventions](#coding-conventions)
-9. [Questions?](#questions)
+9. [Developing in Docker](#developing-in-docker)
```
simon04, line 15: The section is missing in the file (has not been committed)?

.env
```diff
@@ -0,0 +1,2 @@
+USER_ID=1000
+GROUP_ID=1000
```
simon04, line 2: Maybe include a comment `# for Dockerfile-dev`?

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.7'
```
simon04, line 1: Since we already have two `Dockerfile`s for running devdocs in production, I would not expect `docker-compose.yml` to be meant for development only.

Dockerfile-dev
```diff
@@ -0,0 +1,42 @@
...
+    chown --changes --silent --no-dereference --recursive \
+          --from=33:33 ${USER_ID}:${GROUP_ID} \
+        /home/devuser \
+    ;fi
+
+# Host directories cannot be mounted by Dockerfiles so we have to use
+# a temporary directory to hold the lock files while the gems are installed.
+WORKDIR /devdocs/
+
+COPY Gemfile Gemfile.lock Rakefile /devdocs/
```
simon04, line 29: Running `docker build - < Dockerfile-dev` failed:

```
Step 9/14 : COPY Gemfile Gemfile.lock Rakefile /devdocs/
COPY failed: stat /var/lib/docker/tmp/docker-builder202400793/Gemfile: no such file or directory
```

(Not exactly sure how it is meant to be used, though…)

ojeytonwilliams:
Thanks for the review, Simon.  The latest commits should have addressed all your concerns - including adding the missing docs (thanks reflog :sweat:).

ojeytonwilliams:
Hey @simon04, I'm still working on this.  Right now using `thor: upload` doesn't work with AWS, so I need to figure it out.

tsengia:
@ojeytonwilliams and @simon04 Any updates on this? I can pick this up if you'd like. I cloned the repo and started making a devcontainer... and then realized you guys already started!

PR #1351, opened 2020-11-19
Update 5 documentations to their latest versions

samfundev:
Updated async, bluebird, codeception, codeceptjs and electrion to their latest versions.


simon04 (approved):
Awesome, thank you!

assets/javascripts/templates/pages/about_tmpl.coffee
```diff
@@ -258,7 +258,7 @@ credits = [
     'https://api.drupal.org/api/drupal/LICENSE.txt'
   ], [
     'Electron',
-    '2013-2018 GitHub Inc.',
+    'GitHub Inc.',
```
scherepn, line 261: If the copyright year changed, then it should be changed, not dropped. Year of creation information is relevant.

samfundev: I removed it to mirror what they did in the [electron website license](https://github.com/electron/electronjs.org/blob/738c66d8c768fcc2c56e178da0da05a67e266ad5/LICENSE.md). This is different then what the [main electron license](https://github.com/electron/electron/blob/70029aa069d97f3923794b387a7c90ada90189de/LICENSE) does but I went with the website license because that's what's being scrapped. I can change it if people think the main electron license should be used.

PR #1354, opened 2020-11-19
Update Typescript to 4.0.5

MasterEnoc:

### DIFF
--- a/lib/docs/filters/typescript/clean_html.rb
+++ b/lib/docs/filters/typescript/clean_html.rb
@@ -17,6 +17,13 @@ module Docs
           node.name = 'h1'
         end
 
+        css('a:contains("Try")').remove
+        css('pre').each do |node|
+          node.content = node.content
+          node['data-language'] = 'typescript'
+          node.remove_attribute('class')
+        end
+
         # 'Next' title area
         css('.whitespace-tight').remove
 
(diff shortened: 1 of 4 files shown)


lib/docs/filters/typescript/clean_html.rb
```diff
@@ -2,30 +2,38 @@ module Docs
...
+        css('#handbook-content > h2').each do |node|
+          node.name = 'h1'
         end
 
-        css('pre > code').each do |node|
-          node.parent['data-language'] = node['class'].sub('ts', 'typescript').sub('js', 'javascript').remove('language-') if node['class']
-          node.content = node.content.gsub('    ', '  ')
-          node.before(node.children).remove
+        css('a:contains("Try")').remove
+        css('pre').each do |node|
```
simon04, line 21: I've (re-)added a cleanup of the code blocks

simon04 (approved):
Thank you!

lib/docs/scrapers/typescript.rb
```diff
@@ -2,26 +2,37 @@ module Docs
   class Typescript < UrlScraper
     self.name = 'TypeScript'
     self.type = 'simple'
-    self.release = '3.7.4'
-    self.base_url = 'https://www.typescriptlang.org/docs/'
-    self.root_path = 'tutorial.html'
+    self.release = '4.0.5'
+    self.base_url = 'https://www.typescriptlang.org/docs/handbook'
```
scherepn, line 6: This might have been the source of your frustrations with updating this scraper. This page doesn't exist, but does respond with a 301 redirect to `https://www.typescriptlang.org/docs/handbook/`, so it appeared to work.

This is why `root_page?` calls didn't work.

PR #1355, opened 2020-11-19
Update QT documentation

scherepn:

### DIFF
--- a/lib/docs/scrapers/qt.rb
+++ b/lib/docs/scrapers/qt.rb
@@ -103,29 +103,39 @@ module Docs
       Licensed under the GNU Free Documentation License, Version 1.3.
     HTML
 
+    version '5.15' do
+      self.release = '5.15'
+      self.base_url = "https://doc.qt.io/qt-#{self.release}/"
+    end
+
+    version '5.14' do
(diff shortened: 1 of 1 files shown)


lib/docs/scrapers/qt.rb
```diff
@@ -103,29 +103,39 @@ class Qt < UrlScraper
...
+
+    version '5.14' do
+      self.release = '5.14'
+      self.base_url = "https://doc.qt.io/qt-#{self.release}/"
+    end
+
     version '5.13' do
       self.release = '5.13'
-      self.base_url = 'https://doc.qt.io/qt-5.13/'
+      self.base_url = "https://doc.qt.io/archives/qt-#{self.release}/"
```
scherepn, line 118: I don't expect these older versions to have changed much, but regenerating them will fix the links in the footers.

simon04: Fair enough, but scraping additional 5×3000 entries resulting in 5×160MiB is too cumbersome. I'll only upload the new versions 5.14 and 5.15 …

simon04 (approved):
Thank you!

PR #1362, opened 2020-11-20
manifest.json: Change display from `standalone` to `minimal-ui`

sclickk:
Adds a back button to the UI


simon04:
Reference: https://web.dev/add-manifest/#display

simon04 (approved):

PR #1241, opened 2020-06-01
Add PyTorch v1.5.0 documentation

xavieryao:
If you’re adding a new scraper, please ensure that you have:

- [x] Tested the scraper on a local copy of DevDocs
- [x] Ensured that the docs are styled similarly to other docs on DevDocs

- [x] Added these files to the <code>public/icons/*your_scraper_name*/</code> directory:
  - [x] `16.png`: a 16×16 pixel icon for the doc
  - [x] `16@2x.png`: a 32×32 pixel icon for the doc
  - [x] `SOURCE`: A text file containing the URL to the page the image can be found on or the URL of the original image itself

This PR introduces all of PyTorch v1.5.0 Python API docs to DevDocs, excluding notes and language bindings. The categories are identical to the Python API section of [https://pytorch.org/docs/stable/index.html](https://pytorch.org/docs/stable/index.html).

#799 will be solved.

PR #1364, opened 2020-11-20
Support PyTorch v1.6.0+ doc structure

xavieryao:
@simon04 Thanks a lot for reviewing and merging my previous pull request #1241  , but I'm afraid the previous code breaks on the latest PyTorch documentation.

The directory structure has changed since PyTorch v1.6.0+, and the current documents on devdocs.io are not indexed and formatted properly. The structure is messy and some links are invalid.

This patch fixes the issue.

### DIFF
--- a/lib/docs/filters/pytorch/clean_html.rb
+++ b/lib/docs/filters/pytorch/clean_html.rb
@@ -2,9 +2,18 @@ module Docs
   class Pytorch
     class CleanHtmlFilter < Filter
       def call
+        breadcrumbs = at_css('.pytorch-breadcrumbs')
+        type_name = breadcrumbs.css('li')[1].content
+
         @doc = at_css('.pytorch-article')
         # Show katex-mathml nodes and remove katex-html nodes
         css('.katex-html').remove
+
+        # pass type_name to following filters as a new node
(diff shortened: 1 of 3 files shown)


lib/docs/filters/pytorch/clean_html.rb
```diff
@@ -2,9 +2,18 @@ module Docs
   class Pytorch
     class CleanHtmlFilter < Filter
       def call
+        breadcrumbs = at_css('.pytorch-breadcrumbs')
+        type_name = breadcrumbs.css('li')[1].content
+
```
scherepn, line 7: Suggestion:
Rather than sticking a temporary sentinel node within the HTML, instead change the order these two filters are executed. The order of execution for `clean_html` and `entries` filter is set on line 12 of `lib/docs/scrapers/pytorch.rb`

xavieryao: Thanks for the suggestion!

scherepn:
Good updates. If you want, I cleaned up the code a bit more, you can see it here:
https://github.com/freeCodeCamp/devdocs/compare/master...scherepn:pytorch-fix

simon04 (approved):
Thank you!

PR #1372, opened 2020-11-22
Update Pandas to 1.1.4

MasterEnoc:


simon04:
Amazing work! A few notes:

1. On the homepage a big blue man was hunting me down


2. Some spaces are missing for the method parameters


3. Syntax highlighting does not work

MasterEnoc:
Images in the root pages are removed and spaces added where you told me.

This scrappers is very slow and the 'Manual' section has a lot of images, I dind't remove those images because show plots that are needed to understand the code used in those pages. I would like to remove the 'Manual' section because contains a lot of images and does not contain API documentation like the other entries, what do you think @simon04? I kept 'the Manual' section because the previous version have it as well.

Every entry is a page, they are not pointing to any section of the same page, and that is the reason there is not highlight syntax.

Please proceed to review,

simon04 (approved):
Thank you!

PR #1376, opened 2020-11-24
Update Phalcon version 3 and add Phalcon version 4

MasterEnoc:

Note that version 2 was removed because its documentation pages are no longer available.

### DIFF
--- a/lib/docs/filters/phalcon/clean_html.rb
+++ b/lib/docs/filters/phalcon/clean_html.rb
@@ -2,45 +2,24 @@ module Docs
   class Phalcon
     class CleanHtmlFilter < Filter
       def call
-        @doc = at_css('.body')
 
-        if root_page?
-          at_css('h1').content = 'Phalcon'
-        end
+        # left navigation bar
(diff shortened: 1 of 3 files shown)


simon04:
> Note that version 2 was removed because its documentation pages are no longer available.

If feasible, we should archive the old versions in devdocs nevertheless…

Syntax highlighting is broken:

MasterEnoc:
Syntax highlight added and style improved. Proceed to review

simon04:
The methods of a class lost their styles…

MasterEnoc:
This lack of style happened because of the html implemented in the page. Now phalcon docs use 'p' and 'table' tags for methods without any particular id or class, as you can see in the images I'm attaching. Style can be implemented but I have to figure out how since traversing through all the 'p' and 'table' tags would break the style.
image
image

scherepn:
Wow. The new phalcon documentation is truly terrible. It's perhaps the worst documentation I've seen so far.

It may actually be easier to contact the Phalcon team and work with them to improve their documentation than fix it on our end.

PR #1382, opened 2020-11-24
docs/maintainers: update list of maintainers

simon04:


simon04:
@ojeytonwilliams, ping!

Maybe we should distinguish between active and former maintainers?

ojeytonwilliams:
Hey Simon.  Yeah, I like that idea - it's a good way to acknowledge people's efforts, but makes it clear who should and should not be contacted about DevDocs.

simon04:
Maybe change the section heading to "We appreciate major contributions from the following great people"?

ojeytonwilliams:
Tiny nitpick, but I'd say "We appreciate the major contributions made by these great people".  The original sentence sounded a bit like we only wanted PRs from the former maintainers. I tried to make it clear that we appreciate the contributions they have already made.

simon04:
When talking about contributors, we should include additional people from https://github.com/freeCodeCamp/devdocs/graphs/contributors.

Um, complicated. We've the current team (to be contacted via `@freeCodeCamp/devdocs`), former maintainers, and diligent contributors.

ojeytonwilliams:
I think it's fine to just list maintainers and former maintainers.  After all, anyone can go look at the insights tab if they want info about who contributed in general.

simon04:
I've updated the section. @ojeytonwilliams, please review.

ojeytonwilliams (approved):
Perfect, thanks Simon.

PR #1385, opened 2020-11-25
docs/maintainers: remove "Merging pull requests"

simon04:
These guidelines are outdated/obsolete due to CI deployment.

### DIFF
--- a/docs/maintainers.md
+++ b/docs/maintainers.md
@@ -6,7 +6,7 @@ This document is intended for [DevDocs maintainers](#list-of-maintainers).
 
 - PRs should be approved by at least one maintainer before being merged.
 
-- PRs that add or update documentations should always be merged locally, and the app deployed, before the merge is pushed to GitHub.
+- PRs that add or update documentations should always be built and tested locally, and the doc files uploaded by the `thor docs:upload` command, before the PR is merged on GitHub.
 
   This workflow is required because there is a dependency between the local and production environments. The `thor docs:download` command downloads documentations from production files uploaded by the `thor docs:upload` command. If a PR adding a new documentation is merged and pushed to GitHub before the files have been uploaded to production, the `thor docs:download` will fail for the new documentation and the docker container will not build properly until the new documentation is deployed to production.
 


docs/maintainers.md
```diff
@@ -2,14 +2,6 @@
...
 This document is intended for [DevDocs maintainers](#list-of-maintainers).
 
-## Merging pull requests
-
-- PRs should be approved by at least one maintainer before being merged.
-
-- PRs that add or update documentations should always be merged locally, and the app deployed, before the merge is pushed to GitHub.
-
-  This workflow is required because there is a dependency between the local and production environments. The `thor docs:download` command downloads documentations from production files uploaded by the `thor docs:upload` command. If a PR adding a new documentation is merged and pushed to GitHub before the files have been uploaded to production, the `thor docs:download` will fail for the new documentation and the docker container will not build properly until the new documentation is deployed to production.
-
```
ojeytonwilliams, line 12: Correct me if I'm wrong, but isn't 

> If a PR adding a new documentation is merged and pushed to GitHub before the files have been uploaded to production, the `thor docs:download` will fail for the new documentation and the docker container will not build properly until the new documentation is deployed to production.

still true?

simon04: I think we abandoned the "should always be merged locally" part. In contrast, any change should be handled as a GitHub PR. Did I get something wrong?

ojeytonwilliams: > Did I get something wrong?

No, I don't think so.  I was just thinking we should just stress that the docs need uploading before the PR gets merged.  That's right, isn't it?  As I understand it, if you upload the docs *after* merging the PR then they won't be visible on devdocs.io.  Not until the next PR is merged.  

I think that's worth explaining. I'm happy to write the description, if you can confirm I'm not missing something!

simon04: Thanks, I've updated this PR to clarify that merging is performed on GitHub. Is it good to go now, @ojeytonwilliams?

ojeytonwilliams (approved):
Thanks Simon, LGTM :+1:

PR #1387, opened 2020-11-26
Add Julia version 1.5.3

MasterEnoc:

### DIFF
--- a/assets/javascripts/templates/pages/about_tmpl.coffee
+++ b/assets/javascripts/templates/pages/about_tmpl.coffee
@@ -408,7 +408,7 @@ credits = [
     'https://raw.githubusercontent.com/jquery/api.jqueryui.com/master/LICENSE.txt'
   ], [
     'Julia',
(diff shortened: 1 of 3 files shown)


scherepn:
The Julia docs are fairly small (<5MB), it should be easy to add Version 1.3 and 1.4 of the documentation too.

MasterEnoc:
Version 1.4 and 1.3 added. Thank you for your suggestion @scherepn

simon04:
For Julia 1.4 (only), some headings got lost:

scherepn:
The docs HTML structure changed starting w/ 1.4.

I think the full 'correct' way to solve this is to create a new cleaner filter specifically for 1.4+ versions so future maintainers can keep them separate. The pages are substantially similar though, so cheating is probably fine.

https://github.com/scherepn/devdocs/commit/17cbea7d4eea8fe3fb5cdde799ea3a1a4ab53893

simon04 (approved):
Thank you!

PR #1388, opened 2020-11-26
Update Sass to 3.6.4

MasterEnoc:

### DIFF
--- a/lib/docs/filters/sass/entries.rb
+++ b/lib/docs/filters/sass/entries.rb
@@ -1,89 +1,53 @@
 module Docs
   class Sass
     class EntriesFilter < Docs::EntriesFilter
-      TYPES = ['CSS Extensions', 'SassScript', '@-Rules and Directives',
-        'Output Styles']
-
-      SKIP_NAMES = ['Interactive Shell', 'Data Types', 'Operations',
-        'Division and /', 'Keyword Arguments']
-
-      REPLACE_NAMES = {
-        '%foo'               => '%placeholder selector',
-        '&'                  => '& parent selector',
-        '$'                  => '$ variables',
-        '`'                  => '#{} interpolation',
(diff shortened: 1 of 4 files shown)


lib/docs/filters/sass/entries.rb
```diff
@@ -1,89 +1,53 @@
...
       def get_type
-        'Functions'
+        return 'Syntax' if slug.include?('syntax')
+        return 'Style rules' if slug.include?('style')
+        return 'At-Rules' if slug.include?('at')
+        return 'Values' if slug.include?('values')
+        return 'Operators' if slug.include?('operators')
+        return 'Command line' if slug.include?('cli')
+        return 'Modules' if slug.include?('modules')
+        'Misc'
```
scherepn, line 17: This is only a code style thing, but a cleaner way to do this is with this code:
```ruby
case slug
when /syntax/
  'Syntax'
when /style/
  'Style'
...
else
  'Misc'
end
```

MasterEnoc: Thank you!, any suggestion to improve the quality of the code is welcome

lib/docs/scrapers/sass.rb
```diff
@@ -1,31 +1,31 @@
...
-    options[:only] = %w(Sass/Script/Functions.html)
     options[:root_title] = false
     options[:title] = 'Sass Functions'
 
-    options[:container] = ->(filter) do
-      filter.root_page? ? '#filecontents' : '#instance_method_details'
-    end
+    options[:skip_patterns] = [/breaking-changes/]
+
+    # options[:container] = '#main-content'
```
scherepn, line 19: Was this line accidentally left in?

simon04 (approved):
Thank you!

PR #1395, opened 2020-12-01
Update Perl documentation

scherepn:

During development I discovered a problem somewhere with the file handling/page lookup code. There are some pages with filenames that get URL encoded/decoded and the devdocs website doesn't handle them properly. For example, the pages `functions/$#`, `functions/$^`. This is outside my area of understanding and couldn't make much progress identifying a root cause.

Also, the perl website was a bit flaky during testing. I would run the scraper multiple times and occasionally get different results. If any pages show up with the 'Other' type, run it again - there should be no pages with that type.

### DIFF
--- a/assets/javascripts/templates/pages/about_tmpl.coffee
+++ b/assets/javascripts/templates/pages/about_tmpl.coffee
@@ -578,7 +578,7 @@ credits = [
     'https://raw.githubusercontent.com/pydata/pandas/master/LICENSE'
   ], [
     'Perl',
-    '1993-2016 Larry Wall and others',
+    '1993-2020 Larry Wall and others',
     'GPLv1',
     'https://perldoc.perl.org/index-licence.html'
   ], [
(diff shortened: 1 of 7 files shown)


simon04:
For https://perldoc.perl.org/5.28.3/variables/$%23, the documentation is stored as `public/docs/perl~5.28/variables/$%23.html` and accessible via http://localhost:9292/docs/perl~5.28/variables/$%2523.html (`%23` is the URL encoding for `#`, `%25` is the URL encoding for `%`). However, the frontend does not even perform a request for this HTML file. Strange…

Maybe, we can exclude https://perldoc.perl.org/5.28.3/variables/… – those variables are listed here as well https://perldoc.perl.org/5.28.3/perlvar#$%23. And that's where the current https://devdocs.io/perl~5.26/perlvar#%24%23 links to.

scherepn:
Good idea
Dropped the dedicated `functions/` and `variables/` pages for `perfunc` and `perlvar` instead.

simon04 (approved):
Thank you!

PR #1398, opened 2020-12-02
Add HAProxy documentation

scherepn:


simon04:
Amazing work!

I noticed one little glitch, i.e., the anchor links should probably be removed as they link to a non-existing anchor: link is http://localhost:9292/haproxy~2.3/configuration#3.1-cpu-map, but should be http://localhost:9292/haproxy~2.3/configuration#cpu-map



What do you think?

lib/docs/scrapers/haproxy.rb
```diff
@@ -0,0 +1,62 @@
+module Docs
+  class Haproxy < UrlScraper
+    self.type = 'haproxy'
```
simon04, line 3: Maybe we should set `self.name = 'HAProxy'` for a correct typesetting?

scherepn:
Ah, good catch.
The source documents have 5 or 6 different anchors added in for everything record and I worked over this scraper a few times changing which anchors were used multiple times. Looks like I missed a few.

Removing obvious links that go nowhere is easy, I'll fix it.

simon04 (approved):
Thank you!

PR #1399, opened 2020-12-04
Update Relay to 10.1.0

MasterEnoc:

I used the main page of Relay as the base url because index.html refuses being generated using 'https://relay.dev/docs/en/introduction-to-relay/' as base url.

Also note that I only use the API section of Relay for this scraper, if is necessary I can add the other sections.

### DIFF
--- a/lib/docs/filters/relay/clean_html.rb
+++ b/lib/docs/filters/relay/clean_html.rb
@@ -2,58 +2,42 @@ module Docs
   class Relay
     class CleanHtmlFilter < Filter
       def call
-        @doc = at_css('.inner-content, article.withtoc')
 
-        if root_page?
-          at_css('h1').content = 'Relay Documentation'
+        if slug == 'index'
+          css('img').remove
+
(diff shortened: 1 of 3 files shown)


simon04 (approved):
Thank you!

lib/docs/filters/relay/clean_html.rb
```diff
@@ -2,58 +2,42 @@ module Docs
...
-          node['data-language'] = node['class'][/(?<=language\-)(\w+)/]
-          node.content = node.content
-        end
+        css('.edit-page-link').remove
 
-        css('blockquote > p:first-child').each do |node|
-          node.remove if node.content.strip == 'Note:'
+        css('h2, h3').each do |node|
+          node.css('a').remove
+          node['id'] = node.content.gsub(/\s/, '-').downcase
```
scherepn, line 26: As a general rule, try to use `id`s or `name`s from existing anchors/tags rather than creating new ones. Documents usually have existing links and this can interfere/break them.

PR #1400, opened 2020-12-04
Add Groovy documentation

scherepn:


lib/docs/filters/groovy/clean_html.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+
+        css('ul.blockList > table').each do |node|
+          node.parent.before(node).remove
+        end
+
+        css('h3', 'h4').each do |node|
+          node.name = node.name.sub(/\d/) { |i| i.to_i - 1 }
+        end
+
```
simon04, line 82:
```suggestion

        css('pre').each do |node|
          node['data-language'] = 'groovy'
        end
```

simon04 (commented):
Can we link to out Java docs, i.e., replace the link https://docs.oracle.com/javase/8/docs/api/java/io/File.html with https://devdocs.io/openjdk~8/java/io/file?

scherepn:
There's a long history/discussion with internal linking between documentation sets: https://github.com/freeCodeCamp/devdocs/issues/234

If we do decide to do something, I think this would be a good test case.

scherepn:
I just noticed every package link on the homepage is broken.

lib/docs/filters/groovy/clean_html.rb
scherepn: Added. For some reason I thought Prism didn't support Groovy.

simon04 (approved):
Awesome, thank you!

PR #1407, opened 2020-12-07
Update Nim to 1.4.2

MasterEnoc:

## Caveats
There are some entries with duplicate name. This happen because some procedures, macros and templates share the same name, look at the following image:

image

The module 'astalgo' has five 'debug' entries, where three are procedures with different arguments and two are templates. Any idea or suggestion to improve this would be helpful.


simon04 (approved):
Awesome, thank you!

PR #1412, opened 2020-12-07
Fix Bootstrap scraper get_latest_version

scherepn:
The bootstrap team updated their website in preparation for the release of version 5. It's still in beta (and as a rule DevDocs doesn't host beta or prerelease content) so I'll submit the update to the scraper itself once the full 5.0 is available.


simon04 (approved):
Thank you!

PR #1414, opened 2020-12-08
Update Bash to 5.1

MasterEnoc:

### DIFF
--- a/lib/docs/filters/bash/entries.rb
+++ b/lib/docs/filters/bash/entries.rb
@@ -1,13 +1,22 @@
 module Docs
   class Bash
     class EntriesFilter < Docs::EntriesFilter
+
       def get_name
-        name = at_css('hr + a + *').content.gsub(/(\d+\.?)+/, '')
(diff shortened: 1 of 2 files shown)


lib/docs/filters/bash/entries.rb
```diff
@@ -1,13 +1,22 @@
...
     class EntriesFilter < Docs::EntriesFilter
+
       def get_name
-        name = at_css('hr + a + *').content.gsub(/(\d+\.?)+/, '')
+        name = at_css('h1','h2', 'h3', 'h4').content.gsub(/(\d+\.?)+/, '')
+
+        # remove 'E.' notation for appendixes
+        if name.match?(/[[:upper:]]\./)
+          # remove 'E.'
+          name = name.sub(/[[:upper:]]\./, '')
```
scherepn, line 11: You can use the `name.sub!` function to do an in-place operation on a string instead.

MasterEnoc:
Code improved. 
Thanks for your suggestions @scherepn.

simon04 (approved):
Thank you!

Issue #1265, opened 2020-08-22
Protip doesn't adapt to preferences


adi-g15:
After the user has changed his preferences for scrolling, the protip recommended should adapt as well.

Current Preference -> 
![preference](https://user-images.githubusercontent.com/37269665/90957421-0608c500-e4ab-11ea-8397-2a5576d62e1b.png)

Protip suggested ->
![protip](https://user-images.githubusercontent.com/37269665/90957423-086b1f00-e4ab-11ea-94b7-3b7313b4474c.png)

Steps to reproduce->
1. Change your preferences to use arrow keys to scroll main content area
2. Use arrow keys, then a protip popup will likely popup

nshki:
👋🏻 Hey @simon04, I believe this is closable with #1418 and #1431 now merged.

adi-g15:
I closed this, since the pr by @nshki already merged fixing this 🙂

PR #1418, opened 2020-12-10
Update arrow scroll copy

nshki:
# Outcome

This PR makes the arrow scroll copy in the ProTip popup dynamic to reflect proper key combinations depending on if the setting is enabled or not.

#### Settings page

Screenshot_2020-12-09 DevDocs

I updated the copy here to read `shift` instead of `alt` since enabling this setting doesn't actually allow for `alt` with left/right arrow keys. `alt` + left/right results in previous/next pages. `shift` + left/right works properly.

#### ProTip, arrow scroll enabled

Screenshot_2020-12-09 DevDocs(1)

#### ProTip, arrow scroll disabled

Screenshot_2020-12-09 DevDocs(2)

# Issue

#1265.


simon04:
Thanks for spitting this glitch! The user guide at https://devdocs.io/help should be adapted as well, please open a separate PR for this…

simon04 (approved):
Thank you!

PR #1421, opened 2020-12-12
Update Npm to 7.1.2

MasterEnoc:

### DIFF
--- a/lib/docs/filters/npm/clean_html.rb
+++ b/lib/docs/filters/npm/clean_html.rb
@@ -2,36 +2,17 @@ module Docs
   class Npm
     class CleanHtmlFilter < Filter
       def call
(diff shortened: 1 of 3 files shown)


simon04 (commented):
The https://devdocs.io/npm/files/package.json section is missing. Intended?

MasterEnoc:
Yes, I did not add 'package.json' section since it does not add any API information and does not contain configuration variables like 'Config' does.

simon04:
In the past, I often found myself to lookup the specification of certain `package.json` keys. Therefore I would keep it included…

MasterEnoc:
'package.json' section added.

simon04 (approved):
Thank you!

PR #1426, opened 2020-12-13
Update self.links in several scrapers

scherepn:

This is a very small change, I don't think we need to regenerate the doc sets for these now.


simon04 (approved):
Thank you!

PR #1427, opened 2020-12-13
Improve syntax highlight of Sass

MasterEnoc:
Syntax highlight improved in Sass. I did not know how Prism.js worked so I hard coded the syntax highlight, now should be better.


simon04 (approved):
Thank you!

PR #1430, opened 2020-12-14
Fix broken PyTorch scraper

### LINKED ISSUE #1429: Pytorch some docs are missing
On the left side of the pytorch documentation page, you'll see all the libraries - https://pytorch.org/docs/stable/data.html
But torch.utils.data among others are not found in the devdocs data.

These are the only torch.utils.* libraries found in devdocs
image

scherepn:
Fixes #1429


scherepn:
This bug was introduced in commit 5afcd78 🤦

simon04 (approved):
Thank you!

Issue #1404, opened 2020-12-05
Missing parts in TypeScript docs


afsheen1:
I really appreciate the hard work of the TEAM ;) 👍 

I was using the offline version mostly because of no internet connections. but the new update completely changed some of offline module docs. ex( typescript 4.1.2, javascript, ...) 👎 

in type script the hard and most useful part is the configuration file, before new update that was very useful but with new update lots of section changed.. 

ex : before update there was clear section (installation and configuration) for typescript but with new update both of lost. and now it's linked as external link :( also there's some of javascript section linked as external pages.

as I mentioned when someone is using the offline (there would be problem on net connection) so the only parts of doc should use as external link which is rarely using ex in typescript (**Using TypeScript in several environments**) should be external link. that's ok because this is out of offline docs scope.

I hope the new update coverup these missing parts.

thanks :)

MasterEnoc:
@afsheen1 An update for typescript was made hours ago, check if the new documentation is what you requested. If something is missing please let us now in this issue.

afsheen1:
@MasterEnoc Thank you,
Yeah now it's great.. but still one part is displaying as external link **(TSConfig Reference)** and I guess it's the most wanted part of typescript..

as I said in my first comment, after the new update lots of Javascript parts are also missing.

1. Increment & decrement
2. Unary operators
3. Arithmetic operators
4. Relational operators
5. Equality operators
6. Bitwise shift operators
7. Binary bitwise operators
....

I have no idea about the reason of missing these parts.
thanks

MasterEnoc:
Typescript in Devdocs now covers almost everything in the [typescript documentation page](https://www.typescriptlang.org/docs). Maybe I'm missing something or the newest documentation does not cover javascript anymore. 

The only section that explicitly mention 'javascript' is covered by Devdocs. Could you write in this issue the links of the pages/sections you mentioned?
![image](https://user-images.githubusercontent.com/52229226/102017789-9dca4180-3d2e-11eb-9fd1-caa2aea6d561.png)

afsheen1:
Hi  @MasterEnoc 

Thank you, as I said before Yes **(Typescript documentation)** in Devdocs is great now. the only part related to typescript which is still missing from Devdocs is [https://www.typescriptlang.org/tsconfig](url).


for not misunderstanding the following missing parts is related to **(Javascript documentation)** in Devdocs.
[...]

PR #1432, opened 2020-12-17
Typescript: add 'tsconfig' section and update scraper to 4.1.3

MasterEnoc:

Add the documentation requested in #1404 and updates Typescript scraper.

### DIFF
--- a/lib/docs/filters/typescript/clean_html.rb
+++ b/lib/docs/filters/typescript/clean_html.rb
@@ -1,8 +1,16 @@
 module Docs
   class Typescript
     class CleanHtmlFilter < Filter
+
       def call
-        root_page? ? root : other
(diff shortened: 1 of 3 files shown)


simon04 (approved):
Thank you!

scherepn:
This update changes the URLs of all the TypeScript pages. This is generally considered poor form because it messes with search indexes and bookmarks. DevDocs has a mapping function to remap old URLs to new URLs, but it's much easier to just keep the URLs the same if possible.

It is actually possible to add the `tsconfig` file to the docset without changing the old URLs. It also involves less code too. It relies on the`MultipleBaseUrls` module for `UrlScraper`, which isn't typically used and isn't documented anywhere.

Check out PR #1435 to see the code.

PR #1435, opened 2020-12-19
Restore old URLs to TypeScript documentation

scherepn:

The previous update changed the URLs of all the TypeScript pages. This is generally considered poor form because it messes with search indexes and bookmarks. DevDocs has a mapping function to remap old URLs to new URLs, but it's much easier to just keep the URLs the same if possible.

It is actually possible to add the tsconfig file to the docset without changing the old URLs. It doesn't even take much code. This relies on theMultipleBaseUrls module for UrlScraper, which isn't typically used and isn't documented anywhere.

This also improves the syntax highlighting.


simon04 (approved):
Thank you!

PR #1440, opened 2020-12-23
Fix CodeIgniter scraper

scherepn:

This isn't a version update, but a modification that adds some missing entries. The HTML tags changed slightly between v3 and v4 of CodeIgniter and a few entries were missing.


simon04 (approved):
Thank you!

PR #1458, opened 2021-01-04
Add OpenJDK version 15

MasterEnoc:

This is a WIP PR, Openjdk 16 still has some style issues.

All the tables are made with 'div' tags rather than table (th, tr, td, etc) tags, this has been painfully hard to work with and I have not found a good way to fix this.
image

All the other version work as expected and have good styles.

### DIFF
--- a/docs/file-scrapers.md
+++ b/docs/file-scrapers.md
@@ -127,23 +127,16 @@ tar xf ocaml-4.10-refman-html.tar.gz --transform 's/htmlman/ocaml/'
 ```
 
 ## OpenJDK
+Search 'Openjdk' in https://www.debian.org/distrib/packages, find the `openjdk-$VERSION-doc` package,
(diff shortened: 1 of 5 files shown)


simon04:
The resulting OpenJDK 16 documentation is very hard to read (for `Color`):
Screenshot 2021-01-07 at 20 57 12
Screenshot 2021-01-07 at 20 57 28
Screenshot 2021-01-07 at 20 57 46


Should we omit version 16 for now and only add versions 13/14/15?

----

Version 15 could get some syntax highlighting as indicated in this screenshot: 
Screenshot 2021-01-07 at 21 09 11

MasterEnoc:
Version 16 removed and syntax highlight improved for version 15. 

Notice that syntax highlight in the summary tables can lead to undesirable format (at least using prismjs):
image
For this reasons syntax highlight was not added in the summary tables.

simon04:
I've removed versions 13 and 14 since they're end of life and no longer provided on packages.debian.org

simon04 (approved):
Thank you!

PR #1459, opened 2021-01-05
Add Socketio version 2 and 3

MasterEnoc:


simon04 (approved):
Thank you!

scherepn:
This is sort of a subtle thing, but DevDocs assumes the first element of a page fragment is the `h1` tag. This update doesn't remove the enclosing `<div>`s or `<article>`s, so the pages are styled slightly incorrectly.

scherepn:
I haven't tested it, but adding something like:

```ruby
options[:container] = 'article'
```

to the scraper definition file is probably enough.

PR #1460, opened 2021-01-08
Minor Fix PHPUnit, SocketIO

scherepn:
This changes the base HTML tag for page fragments. Fixes small appearance problem on every page with the header.

Compare: 
before
to:
after


simon04 (approved):
Thank you!

PR #1473, opened 2021-01-29
mdn: update scraper to mdn/yari

### LINKED ISSUE #1291: MDN is migrating its documentation to GitHub
MDN is in the process of migrating from its wiki based system to a GitHub/PR based system

- https://github.com/mdn/content
- https://github.com/mdn/content/blob/main/BETA.md
- https://github.com/mdn/yari

simon04:

- [ ] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches it's data in `self.attribution`

Fixes #1291.


lib/docs/filters/mdn/clean_html.rb
```diff
@@ -64,6 +76,9 @@ def call
         end
 
         # New compatibility tables
+        # FIXME(2021):
+        # - fetched from external JSON: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/alignment-baseline/bcd.json
+        # - https://github.com/mdn/yari/blob/master/build/bcd-urls.js
```
simon04, line 81: I've been successfully testing the SVG scraping. 

The **compatibility tables** are the only outstanding issue for now. There are two options (only): (a) leave the compatibility tables unpopulated, (b) for every $mdn_page, try fetching $mdn_page/bcd.json and populate the compatibility tables (this would double the number of fetched URLs).

cc: @ojeytonwilliams

ojeytonwilliams: I see, so the issue is that for each page fetches its own compatibility table rather than having it as part of the html?

It's probably okay to make those extra fetches, if the scraper runs in reasonable time.  I checked the headers for https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/alignment-baseline/bcd.json and found `x-cache: Hit from cloudfront`, so it's not like we'd be putting unreasonable stress on their servers.

simon04: Fetching and rendering the browser-compat-data (bcd) seems somewhat advanced to me. Sending HTTP requests from the ` lib/docs/filters/mdn/clean_html.rb ` does not seem right. When triggering the bcd request in the first place, i don't know how to merge it. This aspect will, unfortunately, block updating MDN docs for a long time (unless someone else steps in)... 😞 

Unfortunately, rendering the compatibility table is disabled for server-side rendering, see https://github.com/mdn/yari/blob/master/client/src/document/lazy-bcd-table.tsx#L70-L72
The rendering itself is also somewhat involved, see https://github.com/mdn/yari/blob/main/client/src/document/ingredients/browser-compatibility-table/index.tsx

ojeytonwilliams: I'd be happy to look into this one, though I'm not sure how long it'll take me (assuming that it's doable).  I think the docs are reasonably up-to-date, though, so I don't think we need to rush and change things.

simon04: I'm happy to hand over this task to you, thank you!

Most MDN docs have been updated in November via https://github.com/freeCodeCamp/devdocs/pull/1313, however, updating DOM has been skipped, see https://github.com/freeCodeCamp/devdocs/pull/1313#issuecomment-727255044 for details. The DOM docs haven't been updated since 2018 (f94dbb0a61999c84f5c621881b8eb4028014a32e) and are highly outdated in some parts.

ojeytonwilliams: @simon04 hey Simon, I know this has slipped a bit.  Would it be okay if we opened this up to contributors? i.e. added a help-wanted flag to the original issue.

I will get to this, if no-one takes it up, but I don't want to give the impression that I'm hard at work on this right now.

simon04: I'd like to unblock updating the DOM docs from implementing/fixing the browser-compat-data scraping. Those docs haven't been updated for >2 years, and thus, the compatibility data is also outdated.

My valuation is that an up-to-date documentation w/o the extra browser-compat-data is more helpful than a >2 year out-of-date documentation/data.

No objections labeling this issue/PR as help-wanted. The other [help-wanted issues](https://github.com/freeCodeCamp/devdocs/issues?q=label%3A%22help+wanted%22) haven't gotten much attention, though. Sorry for not being more optimistic today...

ojeytonwilliams: > My valuation is that an up-to-date documentation w/o the extra browser-compat-data is more helpful than a >2 year out-of-date documentation/data.

I agree entirely.  I'd completely missed the fact that the DOM docs haven't been updated since 2018, sorry about that.  I'll create another issue to track the browser-compat-data problem, but I think you're right that we should carry on updating the docs without it.

simon04: The DOM documentation from the new MDN looks decent. I'll update the DOM docs and merge this PR now in order to unblock dependent PRs...

What a monster the /dom is:

```
  107.9 MiB [##########] public/docs/dom
   18.9 MiB [#         ] public/docs/dom.tar.gz
```

ojeytonwilliams: It does look good and it's nice to have it up to date.

By the way, did you upload a few docs recently?  I noticed that npm, amongst other things, had appeared in the downloads bucket without any changes to the scrapers.  It's not a problem, I was just surprised to see them.

simon04: the doc updates were followed with an update of the scraper versions, for instance see 19d0f58d96ee6986b553002420405c3f3e22d3e5 and  https://github.com/freeCodeCamp/devdocs/commits/main...

ojeytonwilliams: Ah, my mistake. I looked for recently merged PRs, saw nothing relevant and wondered where the docs had been updated.  Thanks for clarifying, I'll check main next time!

PR #1475, opened 2021-02-05
Fix Openjdk additional entries and style

### LINKED ISSUE #1326: OpenJDK 11 scrapper does not pick fragment links
The OpenJDK documentation structure changed slightly between the 8 and 11 releases.  One particular change is that _anchor names_ are no longer used.  Instead, the Javadoc HTML uses _anchor **ids**_.  In other words, the _linkable_ parts of a single page went from:

``` html
<a name="mapToLong-java.util.function.ToLongFunction-">
```

In version 8, to:

``` html
<a id="mapToLong(java.util.function.ToLongFunction)">
```

In version 11.  Note also the use of parenthesis in method identifiers.  Because the current scrapper is not able to identify anchor points in the OpenJDK 11 documentation, the method names are **not** being indexed and inner page links don't work either.

I believe the fix for this would consist of updating [this block][s] with two alternative versions:

- One for version 8, just as it is
- Another one for version 11, replacing `css('a[name]')` with `css('a[id]')`

I currently don't have the expertise to prepare and test a patch, so I'm reporting this just in case someone else can fix it in a shorter time.  However, if nobody else can I'll try to do it myself in the near future.

[s]: /freeCodeCamp/devdocs/blob/master/lib/docs/filters/openjdk/clean_html.rb#L9-L17

MasterEnoc:
Fix #1326 and improve style. 

Now additional entries should be generated in version 11 and 15, improvement in tables styles and add more documentation.

### DIFF
--- a/lib/docs/filters/openjdk/clean_html.rb
+++ b/lib/docs/filters/openjdk/clean_html.rb
@@ -5,14 +5,14 @@ module Docs
   class Openjdk
     class CleanHtmlFilter < Filter
       def call
-        css('.topNav', '.subNav', '.bottomNav', '.legalCopy', 'noscript', '.subTitle').remove
+        css('.topNav', '.subNav', '.bottomNav', '.legalCopy', 'noscript', '.subTitle', 'hr').remove
 
         # Preserve internal fragment links
         # Transform <a name="foo"><!-- --></a><bar>text</bar>
         #      into <bar id="foo">text</bar>
-        css('a[name]').each do |node|
+        css('a[name]','a[id]').each do |node|
           if node.children.all?(&:blank?)
-            node.next_element['id'] = node['name'] if node.next_element
+            node.next_element['id'] = (node['id'] || node['name'])if node.next_element
             node.remove
           end
         end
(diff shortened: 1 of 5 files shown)


simon04 (approved):
Awesome, thank you!

etanol:
When will this go _live_?

ojeytonwilliams:
Hi @etanol, it should have gone live when it was merged.  Are the changes not appearing for you?

etanol:
The #1326 issue seems to have been fixed for OpenJDK 15 but **not** for OpenJDK 11.  Which is strange, because the _anchor style_ was changed in OpenJDK 11 and should be the same as in OpenJDK 15.

ojeytonwilliams:
Ah, I think I understand the problem.  It looks like, though the scrapers have been updated for OpenJDK 11, the updated docs haven't been uploaded for that version.

@simon04 if I scrape those docs and upload them, they'll get picked up the next time we merge a PR, right?

simon04:
@etanol, sorry, my bad. I skimmed over the PR too quickly and did not notice that it also fixes OpenJDK 11. I've updated the release to [11.0.10](https://packages.debian.org/sid/all/openjdk-11-doc) and pushed an update via c9d6e8e0b46dc6fbc1e25e431aa5809832c264e3.

@ojeytonwilliams, yeah, that should work as well...

ojeytonwilliams:
Thanks for fixing that @simon04 and thanks @etanol for letting us know it had slipped through.

etanol:
Fix verified!  All OpenJDK versions how have a more complete index.  Thank you very much for all the help :-)

PR #1480, opened 2021-02-16
feat: migrate to Cloudflare

ojeytonwilliams:
NOTE: this should not be merged immediately, but it could use a second (or third!) look. @MasterEnoc I saw you've been contributing a load of PRs recently, so if you have chance to glance at this, that would be great.

This should update all the references to MaxCDN to either point to a new S3 bucket (for uploads) or to a subdomain that's handled by Cloudflare.  

Previously the cdn.devdocs.io resources were redirected (by Cloudflare) to MaxCDN.  Now they're just served by Cloudflare.  The path is simply CF -> Heroku, but everything is cached on CF.  I don't believe this will cause any serious cache invalidation issues, simply because everything that needs to be updated regularly gets a new hash each update.  Some images don't get cache-busting hashes, but they're only cached for a day so it should not be a huge problem.

The other major change is that an S3 bucket is used for downloads instead of MaxCDN's push zone.  This bucket is behind CF and our proxies.

### DIFF
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -203,7 +203,7 @@ class DocsCLI < Thor
     docs.each do |doc|
       filename = "#{doc.path}.tar.gz"
       print "[S3 bundle] Uploading #{filename}..."
-      cmd = "aws s3 cp #{File.join(Docs.store_path, filename)} s3://devdocs-downloads/bundles/#{filename} --profile devdocs"
+      cmd = "aws s3 cp #{File.join(Docs.store_path, filename)} s3://devdocs-downloads/#{filename} --profile devdocs"
       cmd << ' --dryrun' if options[:dryrun]
       system(cmd)
     end
@@ -337,7 +337,7 @@ class DocsCLI < Thor
 
   def download_doc(doc)
     target_path = File.join(Docs.store_path, doc.path)
-    open "https://downloads.devdocs.io/bundles/#{doc.path}.tar.gz" do |file|
+    open "https://downloads.devdocs.io/#{doc.path}.tar.gz" do |file|
       FileUtils.mkpath(target_path)
       file.close
       tar = UnixUtils.gunzip(file.path)
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -197,8 +197,7 @@ class DocsCLI < Thor
     end
     puts '[S3] Done syncing.'
 
-    # Upload packages to dl.devdocs.io (used by the "thor docs:download" command)
-    # TODO(MIGRATION): replace this with an S3 bucket upload.
+    # Upload packages to downloads.devdocs.io (used by the "thor docs:download" command)
     puts '[S3 bundle] Begin uploading.'
 
     docs.each do |doc|
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -202,7 +202,7 @@ class DocsCLI < Thor
 
     docs.each do |doc|
       filename = "#{doc.path}.tar.gz"
-      print "[S3 bundle] Uploading #{filename}..."
+      puts "[S3 bundle] Uploading #{filename}..."
       cmd = "aws s3 cp #{File.join(Docs.store_path, filename)} s3://devdocs-downloads/#{filename} --profile devdocs"
       cmd << ' --dryrun' if options[:dryrun]
       system(cmd)
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -191,7 +191,7 @@ class DocsCLI < Thor
     puts '[S3] Begin syncing.'
     docs.each do |doc|
       puts "[S3] Syncing #{doc.path}..."
-      cmd = "aws s3 sync #{File.join(Docs.store_path, doc.path)} s3://devdocs-assets/#{doc.path} --delete --profile devdocs"
+      cmd = "aws s3 sync #{File.join(Docs.store_path, doc.path)} s3://devdocs-staging-documents/#{doc.path} --delete --profile devdocs"
       cmd << ' --dryrun' if options[:dryrun]
       system(cmd)
     end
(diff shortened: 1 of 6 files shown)


lib/tasks/docs.thor
```diff
@@ -191,27 +191,23 @@ class DocsCLI < Thor
...
-          sftp.upload! File.join(Docs.store_path, filename), File.join('', 'public_html', filename)
-          print " OK\n"
-        end
-      end
+    # Upload packages to downloads.devdocs.io (used by the "thor docs:download" command)
+    puts '[S3 bundle] Begin uploading.'
+
+    docs.each do |doc|
+      filename = "#{doc.path}.tar.gz"
+      print "[S3 bundle] Uploading #{filename}..."
```
simon04, line 205:
```suggestion
      puts "[S3 bundle] Uploading #{filename}..."
```

For consistency with S3 syncing. Currently, the output is 

```
[S3 bundle] Uploading vue~3.tar.gz...(dryrun) upload: public/docs/vue~3.tar.gz to s3://devdocs-downloads/vue~3.tar.gz
```

lib/tasks/docs.thor
```diff
@@ -191,27 +191,23 @@ class DocsCLI < Thor
     puts '[S3] Begin syncing.'
     docs.each do |doc|
       puts "[S3] Syncing #{doc.path}..."
-      cmd = "aws s3 sync #{File.join(Docs.store_path, doc.path)} s3://devdocs-assets/#{doc.path} --delete --profile devdocs"
+      cmd = "aws s3 sync #{File.join(Docs.store_path, doc.path)} s3://devdocs-documents/#{doc.path} --delete --profile devdocs"
```
simon04, line 194: The permissions are not yet setup? I'm getting an AccessDenied:

```sh
$ bundle exec thor docs:upload --packaged --dryrun
[S3] Begin syncing.
[S3] Syncing vue~3...
fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
[S3] Done syncing.
[S3 bundle] Begin uploading.
[S3 bundle] Uploading vue~3.tar.gz...(dryrun) upload: public/docs/vue~3.tar.gz to s3://devdocs-downloads/vue~3.tar.gz
[S3 bundle] Done uploading.
```

lib/tasks/docs.thor
```diff
@@ -341,7 +337,7 @@ class DocsCLI < Thor
 
   def download_doc(doc)
     target_path = File.join(Docs.store_path, doc.path)
-    open "http://dl.devdocs.io/#{doc.path}.tar.gz" do |file|
+    open "https://downloads.devdocs.io/#{doc.path}.tar.gz" do |file|
```
simon04, line 340: Interestingly, some downloads fail (the failing docs got an update on 2021-01-30):

```sh
$ bundle exec thor docs:download vue
(1/1) Vue.js 3 FAILED (OpenURI::HTTPError: 403 Forbidden)
Done

$ bundle exec thor docs:download vuex
(1/1) Vuex 4 FAILED (OpenURI::HTTPError: 403 Forbidden)
Done

$ bundle exec thor docs:download erlang
(1/1) Erlang 23 FAILED (OpenURI::HTTPError: 403 Forbidden)
Done

$ bundle exec thor docs:download python
(1/1) Python 3.9 OK
Done
```

simon04 (commented):
I've tested a bunch of CLI commands, see my comments. I've tested the the frontend at https://devdocs.in/ without any issues (I cannot judge the correct proxying/caching, though).

lib/tasks/docs.thor
ojeytonwilliams: Ah, yes, good point.  The final thing I wanted to do was to sync the old bucket (which you have been accessing) with the new ones I set up for the migration.  I'll let you know when this is good to go.

lib/tasks/docs.thor
ojeytonwilliams: They weren't, no, and I'm going to review the policies before migrating, but I've attached one that should let you upload.

ojeytonwilliams:
Thanks Simon. I need to do some last sanity checks and sync the buckets, but I think we're nearly there.

I'll pin an issue before I sync things, so that we don't lose anything.

j-f1:
No problems jump out to me after a cursory glance over the changes.

MasterEnoc:
I did some checks (sniff my packets and see where they are being downloaded) while running `thor docs:download` and it is working as intented, all docs are downloaded from `downloads.devdocs.io`; I did the same while loading [devdocs.in](devdocs.in) and all the documents are fetched from `documents.devdocs.io`. 

[devdocs.in](devdocs.in) works pretty good and solves some issues I had with maxCdn. I use a extension to block certain javascript files and  [devdocs.io](devdocs.io) generates random files that are blocked, this does not happen in  [devdocs.in](devdocs.in).
image

ojeytonwilliams:
@j-f1, that's reassuring - thanks for taking a look.

@MasterEnoc interesting, I knew there were a few issues with MaxCDN, but I didn't realise this was one of them.  The routing is a little simpler now and I'm glad it's fixed that problem.

lib/tasks/docs.thor
simon04: The `--dryrun` command succeeds now:

```sh
$ bundle exec thor docs:upload --packaged --dryrun
[S3] Begin syncing.
[S3] Syncing vue~3...
(dryrun) upload: public/docs/vue~3/api/application-api.html to s3://devdocs-documents/vue~3/api/application-api.html
(dryrun) upload: ...
[S3] Done syncing.
[S3 bundle] Begin uploading.
[S3 bundle] Uploading vue~3.tar.gz...
(dryrun) upload: public/docs/vue~3.tar.gz to s3://devdocs-downloads/vue~3.tar.gz
[S3 bundle] Done uploading.
```

simon04 (approved):
Awesome, great work! I'm looking forward to seeing this PR being merged...

PR #1482, opened 2021-02-19
Add overloaded methods to openjdk 11 & 8

### LINKED ISSUE #1409: Java: Not all overloaded methods can be found
When I search the String.replace() method, that's the search result:
image
Clicking on the first result can only index to the first method.
image
There should actually be another overloaded method that is not searchable.
image
_This one is not next to the previous one._
It would be nice to list all of the overloaded methods in the search results.

**And thank you for your team's work, it really helps well.**

MasterEnoc:
Add overloaded methods as additional entries, fix #1409.

### DIFF
--- a/lib/docs/filters/openjdk/entries.rb
+++ b/lib/docs/filters/openjdk/entries.rb
@@ -30,8 +30,15 @@ module Docs
         entries = []
 
         css('.memberNameLink a').each do |node|
-          next if  !(node['href'].match?(/\(/)) # skip non-methods
-          entries << [self.name + '.' + node.content + '()', slug.downcase + node['href']]
+          next unless node['href'].match?(/[-(]/) # skip non-methods
+
+          if (version=='8' || version == '8 GUI' || version == '8 Web')
+            id = node['href'].gsub(/.*#/, '')
+          else
+            id = slug.downcase + node['href']
+          end
+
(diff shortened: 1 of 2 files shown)


lib/docs/filters/openjdk/entries.rb
```diff
@@ -27,34 +27,17 @@ def get_type
...
-                    summary['id'].include?('field') || summary['id'].include?('constant')
-            summary.parent.css('.memberNameLink a').each do |node|
-              name = node.parent.parent.content.strip
-              name.sub! %r{\(.+?\)}m, '()'
-              id = node['href'].remove(%r{.*#})
-              entries[name] ||= ["#{self.name}.#{name}", id]
-            end
-          end.values
+        css('.memberNameLink a').each do |node|
+          next if  !(node['href'].match?(/\(/)) # skip non-methods
```
simon04, line 33:
```suggestion
          next unless node['href'].match?(/[-(]/) # skip non-methods, 8 uses - for (
```

But still, for Java 8 the generated link is http://localhost:9292/openjdk~8/java/lang/stringstring#replace-char-char- (stringstring instead of string) causing a "The page failed to load."

MasterEnoc:
Fixed

simon04 (approved):
Thank you!

PR #1483, opened 2021-02-19
update Ember docs to latest version

yads:
* fix ember parsing to handle current guides/docs

- [X ] Updated the versions and releases in the scraper file
- [X ] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches it's data in `self.attribution`
- [ X] Ensured the icons and the `SOURCE` file in <code>public/icons/*your_scraper_name*/</code> are up-to-date if the documentation has a custom icon
- [ X] Ensured `self.links` contains up-to-date urls if `self.links` is defined
- [X ] Tested the changes locally to ensure:

### DIFF
--- a/assets/javascripts/templates/pages/about_tmpl.coffee
+++ b/assets/javascripts/templates/pages/about_tmpl.coffee
@@ -273,7 +273,7 @@ credits = [
     'https://www.gnu.org/licenses/gpl-3.0.html'
   ], [
     'Ember.js',
-    '2017 Yehuda Katz, Tom Dale and Ember.js contributors',
+    '2020 Yehuda Katz, Tom Dale and Ember.js contributors',
     'MIT',
     'https://raw.githubusercontent.com/emberjs/ember.js/master/LICENSE'
   ], [
--- a/lib/docs/filters/ember/entries.rb
+++ b/lib/docs/filters/ember/entries.rb
@@ -17,15 +17,17 @@ module Docs
       def get_type
         if base_url.host.start_with?('api')
           name = self.name.remove(/ \(.*/)
-          if name.start_with?('@') || name == 'rsvp'
-            'Packages'
-          elsif name == 'Function'
-            'Functions'
+          if name == 'Function'
+            '3. Functions'
+          elsif at_css('h1').content.start_with?('Package')
+            '2. Packages'
           else
-            'Classes'
+            name = name.remove(' (methods)').remove(' (properties)').remove(' (events)')
+            # Reference gets sorted to the top by default, need to have it with other classes so add a zero width space
+            name == 'Reference' ? 'Reference​' : name
           end
         else
-          'Guide'
+          '1. Guide'
         end
       end
 
--- a/lib/docs/filters/ember/entries.rb
+++ b/lib/docs/filters/ember/entries.rb
@@ -2,60 +2,50 @@ module Docs
   class Ember
     class EntriesFilter < Docs::EntriesFilter
       def get_name
-        if base_url.path.start_with?('/api')
-          name = at_css('h1').child.content.strip
-          # Remove "Ember." prefix if the next character is uppercase
-          name.sub! %r{\AEmber\.([A-Z])(?!EATURES)}, '\1'
+        name = at_css('h1').content
+        if base_url.host.start_with?('api')
+          name.gsub!('Package', '')
+          name.gsub!('Class', '')
+          name.strip!
           name << ' (methods)' if subpath.end_with?('/methods')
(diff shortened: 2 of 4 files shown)


lib/docs/scrapers/ember.rb
```diff
@@ -20,13 +22,7 @@ class Ember < UrlScraper
...
     options[:trailing_slash] = false
 
-    options[:container] = ->(filter) do
-      if filter.base_url.path.start_with?('/api')
-        'main article'
-      else
-        'main'
-      end
-    end
+    options[:container] = 'main'
```
simon04, line 25: Unwrap article for api (for reduced content, and to make various CSS selectors effective, such as `._page > h1:first-child`)

```suggestion
    options[:container] = ->(filter) do
      if filter.base_url.host.start_with?('api')
        'main article'
      else
        'main'
      endhost.start_with?('api')
    end
```

lib/docs/filters/ember/entries.rb
```diff
@@ -2,58 +2,39 @@ module Docs
...
         css('section').each_with_object [] do |node, entries|
           next unless heading = node.at_css('h3[data-anchor]')
           next if node.at_css('.github-link').content.include?('Inherited')
-          name = heading.at_css('span').content.strip
 
-          # Give their own type to "Ember.platform", "Ember.run", etc.
-          if self.type != 'Data' && name.include?('.')
-            type = "#{self.name.remove(/ \(.*/)}.#{name.split('.').first}"
-          end
+          name = heading.at_css('span').content.strip
```
simon04, line 37: This is the list I get when searching for "checkbox". Only 3 out of x properties are shown. The entries Checkbox `(events)`, `.` and `.()` all link to /ember/classes/checkbox/methods.

Screenshot_2021-02-19 Ember js Checkbox — DevDocs

assets/javascripts/templates/pages/about_tmpl.coffee
```diff
@@ -273,7 +273,7 @@ credits = [
     'https://www.gnu.org/licenses/gpl-3.0.html'
   ], [
     'Ember.js',
-    '2017 Yehuda Katz, Tom Dale and Ember.js contributors',
+    'Copyright (c) 2020 Yehuda Katz, Tom Dale and Ember.js contributors',
```
simon04, line 276: We normally omit the "copyright (c)" for brevity.

```suggestion
    '2020 Yehuda Katz, Tom Dale and Ember.js contributors',
```

lib/docs/filters/ember/entries.rb
simon04, line 37: Below "Classes", a gazillion non-working "Function" entries is shown:

Screenshot_2021-02-19 Ember js Classes — DevDocs

simon04 (changes requested):
Hi, thank you for this pull request. I've added a few notes and suggestions. Most importantly, the "Classes" entries need some tuning.

lib/docs/scrapers/ember.rb
yads: The issue with unwrapping the content, is that not all pages in the Ember guides have an article. I originally wanted to have this be 'main article', but as it turned out not all pages have an article tag. There isn't much of a pattern, but I'll see if I can tune it a bit.

lib/docs/filters/ember/entries.rb
yads: The reason that this is only listing items that do not have Inherited in one of the documentation fields (it's actually in the line above this one.

`          next if node.at_css('.github-link').content.include?('Inherited')`

I can get rid of that rule if that helps. I'll investigate the weird `.` and `.()` entries.

yads: Ok I fixed the `Checkbox.` issue, however, it seems like `Checkbox.()` is a bug with the Ember docs themselves. This entry is missing a name https://api.emberjs.com/ember/3.25/classes/Checkbox/methods?anchor= and is thereby coming out like that. I think I'd rather fix it in the Ember docs themselves than have a special case for this in the scraper.

yads: Fixed these and moved them into their own type

yads:
@simon04 I think I've addressed most of the issues you've raised. I've also fixed the broken Packages as the links did not work properly.

lib/docs/scrapers/ember.rb
yads: I wound up reversing the original container rule as it works with the new docs better.

simon04 (approved):
Awesome, thank you!

PR #1486, opened 2021-02-20
Added Documentation for React-bootstrap

dohsimpson:


lib/docs/filters/react_bootstrap/clean_html.rb
```diff
@@ -0,0 +1,17 @@
...
+  class ReactBootstrap
+    class CleanHtmlFilter < Filter
+      def call
+        css('.flex-column.d-flex').remove
+        css('header').remove
+        css('.bs-example').remove
+        css('.position-relative').each do |node|
+          code = node.at_css('textarea')
+          code.name = 'pre'
+          code['style'] = code['style'] + '; border: solid 1px;'
```
simon04, line 11: - [ ] The `.position-relative` does not seem very specific for a code fragment. What about using `css('pre').ancestors('.position-relative').first`?
- [ ] Please extract the raw JSX code from the code fragment and rely on DevDoc's Prism.js syntax highlighting. See other scrapers for some examples. Most of them follow a patter such as: 

```ruby
node.content = node.content
node['data-language'] = 'jsx'
```

lib/docs/scrapers/react_bootstrap.rb
```diff
@@ -0,0 +1,34 @@
...
+    options[:skip] = %w(
+      react-overlays/
+    )
+
+    options[:replace_paths] = {
+    }
+
+    options[:trailing_slash] = true
+
+    options[:attribution] = <<-HTML
```
simon04, line 24: Please also add the license to `assets/javascripts/templates/pages/about_tmpl.coffee`

simon04 (changes requested):
Thank you. Please see my review comments.

lib/docs/filters/react_bootstrap/clean_html.rb
```diff
@@ -0,0 +1,17 @@
+module Docs
+  class ReactBootstrap
+    class CleanHtmlFilter < Filter
```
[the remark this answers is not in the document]
dohsimpson: Hi @simon04 , thanks for the review, I have fixed the second bullet point, which is to remove most of the unnecessary classes. But I would like to keep the styling as "simple" please. As the default "simple" stylesheet looks pretty good in my opinion.

simon04 (approved):
Thank you!

PR #1490, opened 2021-03-03
Fix sidebar when the mouse leaves the window

### LINKED ISSUE #1243: sidebar only stays hidden if mouse leaves to left
In #829 and #831 the sidebar was changed to stay hidden (considering that setting was set) if the mouse left the window, but only if it exited the window from the left edge. I don't think that this makes sense; the sidebar should remain hidden regardless of how the mouse left the window.

MasterEnoc:
Fix #1243. Now the sidebar is hidden whenever the mouse leaves the window, regardless the side from the mouse left the window.


simon04 (approved):
Thank you!

Issue #234, opened 2015-07-15
Intercept links to known documentation


gpgekko:
I wonder if it would be possible to intercept (either during scraping or when clicked) links to known documentation. For example, the Backbone documentation contains quite a few links to Underscore documentation pages. It would be nice if these would open within DevDocs, since both are available.

This would be an improvement for offline use, and would be a nice enhancement for online use too.

Thibaut:
Thanks for the suggestion. This is quite a difficult thing to implement but I agree it would be a nice enhancement. I have no time to work on it at the moment (given my long list of todos and other high-impact features), but maybe in the future.

Hipska:
I think this could also solve #36, #41, #71, #89 and #187.

orionlee:
#514 is also a specific case for this request.

JesseTG:
What would make this difficult, exactly?

Thibaut:
See my comments in https://github.com/Thibaut/devdocs/issues/524

JesseTG:
What if, while parsing server-side, we maintained a map of old links to transformed links, and then exposed this map to other docs?

Thibaut:
@JesseTG that would be quite complex and difficult to maintain. I agree fixing/improving this would be great, but if it comes at the cost of increased maintenance burden (I'm already maxed out) and dependencies between docs/scrapers, then I don't think it's a good trade-off.

Kaijiro:
I think you can apply this with enabled docs, not only installed.

j-f1:
@Kaijiro 

> I think you can apply this with enabled docs, not only installed.

Good point! Fixed.

PR #1495, opened 2021-03-05
Add external_urls filter

MasterEnoc:
This a tiny filter whose purpose is replace urls found in `<a>` tags for relative URLs pointing to devdocs documentation. This is done manually, each author or maintainer should add (if needed) `options[:external_url]` filter option to the scraper.

This is not done server-side as discussed in #234 but it gives an option to avoid links of external documentation when it is available in devdocs. 

This commit already implements this filter in the `backbone.rb` file.

### DIFF
--- a/docs/filter-reference.md
+++ b/docs/filter-reference.md
@@ -84,6 +84,7 @@ The `call` method must return either `doc` or `html`, depending on the type of f
 * [`AttributionFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/attribution.rb) — appends the license info and link to the original document
 * [`TitleFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/title.rb) — prepends the document with a title (disabled by default)
 * [`EntriesFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/entries.rb) — abstract filter for extracting the page's metadata
+* [`ExternalUrlsFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/external_urls.rb) — replaces external URLs for relative URLs of existant devdocs documentation.
 
 ## Custom filters
 
--- a/lib/docs/filters/core/external_urls.rb
+++ b/lib/docs/filters/core/external_urls.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Docs
+  class ExternalUrlsFilter < Filter
+
+    def call
+      if context[:external_urls]
+
+        root = path_to_root
+
+        css('a').each do |node|
+
+          next unless anchorUrl = node['href']
+
+          # avoid links already converted to internal links
+          next if anchorUrl.match?(/\.\./)
+
+          if context[:external_urls].is_a?(Proc)
+            node['href'] = context[:external_urls].call(anchorUrl)
+            next
+          end
+
+          url = URI(anchorUrl)
+
+          context[:external_urls].each do |host, name|
+            if url.host.to_s.match?(host)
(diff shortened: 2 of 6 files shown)


docs/filter-reference.md
```diff
@@ -84,6 +84,7 @@ The `call` method must return either `doc` or `html`, depending on the type of f
 * [`AttributionFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/attribution.rb) — appends the license info and link to the original document
 * [`TitleFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/title.rb) — prepends the document with a title (disabled by default)
 * [`EntriesFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/entries.rb) — abstract filter for extracting the page's metadata
+* [`ExternalUrlsFilter`](https://github.com/freeCodeCamp/devdocs/blob/master/lib/docs/filters/core/external_urls.rb) — replaces external URLs for relative URLs of existant devdocs documentation.
```
simon04, line 87: s/existant/existing

lib/docs/filters/core/external_urls.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Docs
+  class ExternalUrlsFilter < Filter
+
+    def call
+      if context[:external_urls]
+
+        root = path_to_root
```
simon04, line 9: Can we simply use the following?

```suggestion
        root = '/'
```

lib/docs/filters/core/external_urls.rb
```diff
@@ -0,0 +1,38 @@
...
+          if context[:external_urls].is_a?(Proc)
+            node['href'] = context[:external_urls].call(anchorUrl)
+            next
+          end
+
+          url = URI(anchorUrl)
+
+          context[:external_urls].each do |host, name|
+            if url.host.to_s.match?(host)
+              node['href'] = root + name + url.path.to_s + '#' + url.fragment.to_s
```
simon04, line 27: The generated link is correct, but does not work in devdocs. Presumably a JavaScript call to `app.enableDoc(...)` is required.


https://user-images.githubusercontent.com/782446/111040714-acf78300-8434-11eb-8879-74d8412c9f5a.mov

simon04 (commented):
Great idea! Please see my comments.

See also: former PR https://github.com/freeCodeCamp/devdocs/pull/714 by @j-f1

MasterEnoc:
Code improved by your suggestions @simon04. 

Reading the js of devdocs I found that each link ('a' tag) is verified if it is part of devdocs or not before send me to the page that it is pointing, if the link is part of devdocs it will make an xhr request and it will show me content retrieved by the xhr request, this behaviour works if the link is part of the same documentation or if the link is from a documentation that is enabled, if not it will show me an error. I changed this behaviour, if a link is part of devdocs but it is from a different doc it will send to devdocs reloading the page rather than make an xhr request.

assets/javascripts/lib/page.coffee
```diff
@@ -178,16 +178,22 @@ onclick = (event) ->
...
+
   return
 
 isSameOrigin = (url) ->
   url.indexOf("#{location.protocol}//#{location.hostname}") is 0
 
+isSameOriginDifferentDoc = (url) ->
+  console.log(url.pathname)
+  console.log(location.pathname)
+  url.pathname == location.pathname
```
simon04, line 195: For https://devdocs.io/typescript/declaration-files/do-s-and-don-ts#function-overloads, `location.pathname` is `"/typescript/declaration-files/do-s-and-don-ts"`.

Remove the `console.log` statements?

MasterEnoc: `console.log` removed and now should work checking only the initial path of the pathname.

PR #1510, opened 2021-03-29
fix: set branch to master in deploy action

ojeytonwilliams:
$default-branch did not seem to catch merges to master, so this PR names the branch that should trigger the action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,8 @@ name: Deploy
 
 on:
   push:
-    branches: [ $default-branch ]
+    branches:
+      - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -2,7 +2,8 @@ name: Deploy
 
 on:
   push:
-    branches: [ $default-branch ]
+    branches:
+      - master
```
RandellDawson, line 6: @ojeytonwilliams This should be changed to `main`.

ojeytonwilliams: Agreed.  I'll do that and update the action.

PR #1512, opened 2021-03-30
fix: prevent Deploy action running on forks

ojeytonwilliams:
While it would not deploy when a fork updates, due to lack of an api key, it is inefficient. Also, it's potentially confusing to see a failing check on your fork when you haven't broken anything.

I've tested it on my own fork: https://github.com/ojeytonwilliams/devdocs/commit/c2b38c19fa6e9392f2837f75f57c5a0ed8875a38

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-20.04
-
+    if: github.repository == 'freeCodeCamp/devdocs'
     steps:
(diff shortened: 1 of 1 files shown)


simon04 (approved):

ojeytonwilliams:
Thanks, Simon, let's give this a go.

By the way, I'm just setting up the documentation report action now, so that should be in place in time.

ojeytonwilliams:
FYI, it was working, but I manually cancelled the deployment since there was nothing new to deploy.  Hence the failure.

PR #1513, opened 2021-03-30
feat: add doc report scheduled action

ojeytonwilliams:
This should handle the report generation that was previously dealt with by Travis.

### DIFF
--- a/.github/workflows/schedule-doc-report.yml
+++ b/.github/workflows/schedule-doc-report.yml
@@ -1,0 +1,17 @@
+name: Generate documentation versions report
+on:
+  schedule:
+    - cron: '17 4 1 * *'
+
+jobs:
+  report:
+    runs-on: ubuntu-20.04
+    if: github.repository == 'freeCodeCamp/devdocs'
+    steps:
+    - uses: actions/checkout@v2.3.4
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1.66.1
+      with:
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Generate report
+      run: bundle exec thor updates:check --github-token ${{ secrets.DEVDOCS_BOT_PAT }} --upload


.github/workflows/schedule-doc-report.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-20.04
+    if: github.repository == 'freeCodeCamp/devdocs'
+    steps:
+    - uses: actions/checkout@v2.3.4
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1.66.1
+      with:
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+    - name: Generate report
+      run: bundle exec thor updates:check --github-token ${{ secrets.DEVDOCS_BOT_PAT }} --upload
```
simon04, line 17:
```suggestion
      run: bundle exec thor updates:check --github-token ${{ secrets.GITHUB_TOKEN }} --upload
```

The `GITHUB_TOKEN` _should_ be provided automatically, see https://docs.github.com/en/actions/reference/authentication-in-a-workflow

ojeytonwilliams: It should be, yes, but it comes with a set of permissions we don't need: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

`DEVDOC_BOT_PAT` just has `public_repo` access:  https://docs.github.com/en/developers/apps/scopes-for-oauth-apps#available-scopes, which I think is less than what `GITHUB_TOKEN` has.  It's not quite as granular as I would like, but I think it's a bit better.  What do you think?

simon04: Sounds reasonable. Then let's stick with `DEVDOC_BOT_PAT` :)

ojeytonwilliams: Okay, cool.  Thanks again for the speedy reviews, I really appreciate that you're sanity checking these things!

PR #1514, opened 2021-03-30
Add support of compatibility tables

MasterEnoc:
Add support for compatibility tables. I tested with all the mdn scrapers and works with all the scrapers, just one error is raised with the XSLT documentation but this needs to be fixed by mdn (One page does not have index.json file).

I started this PR working with the files of the mdn BCD (Browser Compatibility Data) but this was difficult to work with since there is no way to match a page with its json file at all, using the slug of the page is a good guess but didn't work with all pages, needing a lot of manual intervention. Now it works making (at least two) http requests for page to the mdn server. 

## Test the compatibility tables
Just run the scrapers.

## Issues  
Making http request is extremely slow, it took me more than an hour to scrap all the `dom` documentation. I think this could be solved adding threads.


lib/docs/filters/html/clean_html.rb
```diff
@@ -6,6 +6,15 @@ def call
           node.before(node.children).remove
         end
 
+        if at_css('#browser_compatibility') and not at_css('#browser_compatibility').next_sibling.classes.include?('warning')
+          at_css('#browser_compatibility').next_sibling.remove
+          begin
+            compatibility_table = generateCompatibilityTable('html', slug, doc)
+            at_css('#browser_compatibility').add_next_sibling(compatibility_table)
+          rescue
+          end
```
MasterEnoc, line 15: These are the debugging statements mentioned in the PR body, some pages still raise errors, I'm still trying to fix them.

lib/docs/core/compat_tables.rb
```diff
@@ -0,0 +1,165 @@
...
+    'chrome_android' => 'Chrome Android',
+    'firefox_android' => 'Firefox for Android',
+    'opera_android' => 'Opera Android',
+    'safari_ios' => 'Safari on IOS',
+    'samsunginternet_android' => 'Samsung Internet'
+  }
+
+  GLOBALFILES = ['global_attributes']
+
+  def generateCompatibilityTable(doc, slug, documentation)
```
MasterEnoc, line 25: I tried to make this somewhat abstract and not use any specific data from the scraper or the filters, this could be helpful if any other documentation needs this tables. However, this could be more abstract but my idea was not make an API of this, was just make html tables with the compatibility data.

lib/docs/core/compat_tables.rb
```diff
@@ -0,0 +1,165 @@
...
+    @documentation = documentation
+    slug.downcase!
+    doc.downcase!
+
+    pathToJson = slug.split('/')
+
+    @tagOrElement = pathToJson.last
+
+    jsonFileString = "browser-compat-data/#{doc}"
```
simon04, line 34:
```suggestion
    jsonFileString = "docs/browser-compat-data/#{doc}"
```

Using the same directory as the file scrapers do.

lib/docs/core/compat_tables.rb
simon04, line 25:
```suggestion
  def generate_compatibility_table(doc, slug, documentation)
```

Ruby seems to follow the snake_case convention for function names.

lib/docs/core/compat_tables.rb
```diff
@@ -0,0 +1,165 @@
...
+    elsif version_added == false
+      version_added = 'No'
+    else
+    end
+
+    if version_removed
+      formatString = "<td>#{version_added}-#{version_removed}"
+    else
+      formatString = "<td>#{version_added}"
+    end
```
simon04, line 150: The green/red coloring for the compatibility tables in MDN is nice ...

simon04 (commented):
Wow, very nice work. Great idea to clone the Git repository locally and use it pretty much as the file scrapers do. I've left a few comments...

I've successfully generated the html docs (with the minor flaws you mentioned) by invoking the following commands:

```sh
gh pr checkout 1473
git clone https://github.com/mdn/browser-compat-data.git
curl https://patch-diff.githubusercontent.com/raw/freeCodeCamp/devdocs/pull/1514.diff | git apply
bundle exec thor docs:generate --force --package html
```

ojeytonwilliams:
I agree with Simon, this is great work.  After following the steps (thanks for the reminder about .patch, by the way!), I was able to get it up and running nice and easily.

There was an error during scraping, though:

```
File does not exist: browser-compat-data/html/elements/heading_elements
ERROR:                                                                                                                                                                                                     
  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl
  NoMethodError: undefined method `content=' for nil:NilClass

  ../devdocs/lib/docs/filters/mdn/clean_html.rb:45:in `block in call'
  ../devdocs/lib/docs/filters/mdn/clean_html.rb:44:in `call'
  ../devdocs/lib/docs/core/instrumentable.rb:15:in `instrument'
  ../devdocs/lib/docs/core/instrumentable.rb:15:in `instrument'
  ../devdocs/lib/docs/core/scraper.rb:179:in `process_response'
  ../devdocs/lib/docs/core/scraper.rb:269:in `process_response'
  ../devdocs/lib/docs/core/scraper.rb:160:in `block in handle_response'
  ../devdocs/lib/docs/core/instrumentable.rb:15:in `instrument'
  ../devdocs/lib/docs/core/scraper.rb:159:in `handle_response'
  ../devdocs/lib/docs/core/scraper.rb:77:in `block in build_pages'
  ../devdocs/lib/docs/core/requester.rb:59:in `block (2 levels) in handle_response'
  ../devdocs/lib/docs/core/requester.rb:58:in `each'
  ../devdocs/lib/docs/core/requester.rb:58:in `block in handle_response'
  ../devdocs/lib/docs/core/instrumentable.rb:15:in `instrument'
  ../devdocs/lib/docs/core/requester.rb:57:in `handle_response'
  ../devdocs/lib/docs/core/requester.rb:18:in `run'
  ../devdocs/lib/docs/core/scrapers/url_scraper.rb:38:in `request_all'
  ../devdocs/lib/docs/core/scraper.rb:76:in `build_pages'
  ../devdocs/lib/docs/core/scraper.rb:241:in `fetch_internal_urls'
  ../devdocs/lib/docs/core/scraper.rb:232:in `with_internal_urls'
  ../devdocs/lib/docs/core/scraper.rb:223:in `store_pages'
  ../devdocs/lib/docs.rb:100:in `generate'
  ../devdocs/lib/tasks/docs.thor:306:in `generate_doc'
  ../devdocs/lib/tasks/docs.thor:106:in `generate'
```
and http://localhost:9292/html/element/dl never got generated.

As for the issues you raised, it seems fine if this PR is only a partial solution. Particularly for pages like https://developer.mozilla.org/en-US/docs/Web/HTML/Microdata, where the compatibility data is a bit lacking anyway.

I would defer to @simon04's judgment on this, but as long as all the pages we want exist, I think it's fine if some of them are missing compatibility data.  At least for this PR, we can always fix it later.

ojeytonwilliams:
Hey @MasterEnoc would you like us to take another look or are you still refining things?

MasterEnoc:
Im still improving the code, I will let you know when you can give a try again

MasterEnoc:
@simon04 @ojeytonwilliams You can try again this PR. Just check the PR body again, I updated it after changing how the compatibility tables work.

simon04:
Wow, that's great news!

## Rendering

Very nice :smile:. I've tested https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle – there seems to be an issue for extracting the versions?

- MDN: Screenshot_2021-05-04  circle - SVG Scalable Vector Graphics MDN
- current DevDocs: 
Screenshot_2021-05-04 SVG circle — DevDocs(1)
- this PR: 
Screenshot_2021-05-04 SVG circle — DevDocs

## Performance 

Scraping svg takes a while, likely due to the HTTP requests made by `compat_tables.rb`? I wouldn't mind cloning the Git repository https://github.com/mdn/browser-compat-data to `docs/`

## Code

What about turning `compat_tables.rb` into a separate filter?

```patch
diff --git a/lib/docs/core/compat_tables.rb b/lib/docs/core/compat_tables.rb
deleted file mode 100644
index 82fbac31..00000000
--- a/lib/docs/core/compat_tables.rb
+++ /dev/null
@@ -1,194 +0,0 @@
-module CompatTables
 ...
-end
diff --git a/lib/docs/filters/mdn/compat_tables.rb b/lib/docs/filters/mdn/compat_tables.rb
new file mode 100644
index 00000000..3bec5763
--- /dev/null
+++ b/lib/docs/filters/mdn/compat_tables.rb
@@ -0,0 +1,213 @@
+module Docs
+  class Mdn
+    class CompatTablesFilter < Filter
+
+      def call
+        if at_css('#browser_compatibility') \
+            and not at_css('#browser_compatibility').next_sibling.classes.include?('warning') \
+            and not at_css('#browser_compatibility').next_sibling.content.match?('Supported')
+  
+            at_css('#browser_compatibility').next_sibling.remove
+  
+            compatibility_tables = generate_compatibility_table()
+            compatibility_tables.each do |table|
+              at_css('#browser_compatibility').add_next_sibling(table)
+            end
+          end  
+        doc
+      end
+
+      def generate_compatibility_table()
 ... paste lib/docs/filters/mdn/compat_tables.rb
diff --git a/lib/docs/core/filter.rb b/lib/docs/core/filter.rb
index 8f6bf173..709049bf 100644
--- a/lib/docs/core/filter.rb
+++ b/lib/docs/core/filter.rb
@@ -1,12 +1,8 @@
 # frozen_string_literal: true
 
-require './lib/docs/core/compat_tables.rb'
-
 module Docs
   class Filter < ::HTML::Pipeline::Filter
 
-    include CompatTables
-
     def css(*args)
       doc.css(*args)
     end
diff --git a/lib/docs/filters/svg/clean_html.rb b/lib/docs/filters/svg/clean_html.rb
index 15904229..d0acf6fe 100644
--- a/lib/docs/filters/svg/clean_html.rb
+++ b/lib/docs/filters/svg/clean_html.rb
@@ -26,18 +26,6 @@ module Docs
           node['class'] = 'index'
           css('h3').each { |n| n.name = 'span' }
         end
-
-        if at_css('#browser_compatibility') \
-          and not at_css('#browser_compatibility').next_sibling.classes.include?('warning') \
-          and not at_css('#browser_compatibility').next_sibling.content.match?('Supported')
-
-          at_css('#browser_compatibility').next_sibling.remove
-
-          compatibility_tables = generate_compatibility_table()
-          compatibility_tables.each do |table|
-            at_css('#browser_compatibility').add_next_sibling(table)
-          end
-        end
       end
 
     end
diff --git a/lib/docs/scrapers/mdn/mdn.rb b/lib/docs/scrapers/mdn/mdn.rb
index 03a1d44b..2b31668a 100644
--- a/lib/docs/scrapers/mdn/mdn.rb
+++ b/lib/docs/scrapers/mdn/mdn.rb
@@ -3,7 +3,7 @@ module Docs
     self.abstract = true
     self.type = 'mdn'
 
-    html_filters.push 'mdn/clean_html'
+    html_filters.push 'mdn/clean_html', 'mdn/compat_tables'
 
     options[:container] = '#content > .main-page-content'
     options[:trailing_slash] = false

```

MasterEnoc:
Now compatibility tables is a filter only for the mdn documentation and the problem with numbers in the tables is fixed, thank you for the idea @simon04.

My first idea was to use the mdn BCD repo, but after using it and spent some time coding around it I noticed that It needed a lot of manual intervention. The biggest issue is that there is no good way to find a path to the json file within a page, and I tried to use the `slug` as a clue to find the json file of each page but it didn't work with all the pages, some pages use a json file whose path differ to its slug.

Example: 
slug = [SVG/Attribute/yChannelSelector](https://developer.cdn.mozilla.net/en-US/docs/Web/SVG/Attribute/yChannelSelector)
path to json file = [svg/elements/feDisplacementMap.json](https://github.com/mdn/browser-compat-data/blob/main/svg/elements/feDisplacementMap.json) (One element within this json file is YChannelSelector)

As you can see, the slug is nothing alike the path to the json file. A lot of pages has this problem and I manually intercept the slug and and convert it to the path to the json file, this was a big hash that was increasing its size while I was finding more pages with the previously mentioned problem.

simon04 (approved):
Very nice! Are you going to scrape all the docs and merge this change? (I'm happy to help.)

MasterEnoc:
I just modify the scrapers for the following reasons:

- xslt_xpath raised an error while scraping one specific page, this was due a redirection problem. Now it is fixed.
- There were duplicated titles, I just remove the 'title' filter in the scrapers.
image

MasterEnoc:
@simon04 I already have the xslt and svg documentation, I'm going to scrape the DOM documentation. You can help me scraping Javascript, Html and CSS docs.

ojeytonwilliams:
Congrats on figuring out how to include these @MasterEnoc the docs look great now!

It's a relief that, with the work you and @simon04 have been putting in, the default docs can finally be updated again. 🎉

Issue #786, opened 2018-04-06
Bug Button


alexch:
We already have https://trello.com/b/6BmTulfx/devdocs-documentation for people to request new doc sets, or report updates to existing doc sets, and GitHub issues to report bugs and glitches. Would it also be useful to add a "Report Bug" button to the client site, to encourage people to point out issues even more? It would redirect to a new GitHub Issue page with a bug report template filled in.

@QuincyLarson calls this "Crowdsourcing the [canary](https://en.wiktionary.org/wiki/canary_in_a_coal_mine)" :-)

PR #1524, opened 2021-04-21
Add 'report a bug' button

MasterEnoc:
Add 'report bug' button requested in #786.

### DIFF
--- a/assets/stylesheets/components/_header.scss
+++ b/assets/stylesheets/components/_header.scss
@@ -62,7 +62,7 @@
   top: .25rem;
   right: .25rem;
   width: 8.5rem;
-  height: calc(13.75rem + 1px);
+  height: calc(2.25rem  * 6 + 2.5rem + 1px); // (height of each menu element * total menu elements + menu title element total height + menu title border size)
   white-space: nowrap;
   word-wrap: normal;
   overflow-wrap: normal;
--- a/views/app.erb
+++ b/views/app.erb
@@ -25,6 +25,7 @@
       <a href="/news" class="_menu-link">Changelog</a>
       <a href="/help" class="_menu-link">Guide</a>
       <a href="/about" class="_menu-link">About</a>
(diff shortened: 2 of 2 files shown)


ojeytonwilliams (approved):
LGTM :+1:

simon04 (approved):
Just to mention: GitHub supports passing a custom issue body, e.g. using https://github.com/freeCodeCamp/devdocs/issues/new?labels=bug&body=foo+bar – this way we could automatically prefill the user agent and/or page where the error occurred.

ojeytonwilliams:
Oh, that's neat.  We could include that in this PR if it's a simple tweak (I don't know erbs well enough, honestly).  Otherwise, that could be a follow up.

MasterEnoc:
It could be fine, but the custom issue body will bypass the selection of an issue template; we have two types of bugs templates, doc style and bug report (UI, devdocs features bug, etc), both have different issue body. If it is fine bypass the selection of an issue template I will continue adding the information @simon04 mentioned.

j-f1:
Note that the new issue page (https://github.com/freeCodeCamp/devdocs/issues/new/choose) redirects to a login page if users don’t have a GitHub account. Maybe it would make sense to make a custom support page on devdocs.io that links to each template individually and suggests users create a GH account if they don’t have one yet? I guess it’s true that a DevDocs user is more likely to have a GH account than a random person, though.

MasterEnoc:
Make a new page would end up suggesting the user to make a GH account, right? Unless the page can make anonymous issues (It would be helpful for those who do not want a GH account) I do not find it very useful, I have seen how other bug trackers let make issues via email, I do not know if Github allows that.

j-f1:
Yeah — after adding a GitHub issue link to some Mac apps I wrote I realized that it isn’t really clear why GH would ask you to log in, so I changed the apps to link to a page with my email on it too.

PR #1526, opened 2021-04-22
fix: remove reference to dom_events

ojeytonwilliams:
Since the MDN scrapers have been updated (here: https://github.com/freeCodeCamp/devdocs/commit/a0776af91320fdbeb6e00e115fea8b77153ae42c) this no longer exists.

I noticed the problem when I tried to `thor docs:download --default` (everything else seemed fine).


simon04 (approved):
Thank you!

PR #1536, opened 2021-04-30
Harmonized CLI command version notation

### LINKED ISSUE #1519: Teach `docs:generate` about `@version` notation, consistent with `docs:download`
## Summary

To make working with devdocs more intuitive, it would be helpful if the `docs:*` sub-commands were consistent with respect to `version` handling.

`thor docs:download ruby@2.7` is probably most frequently used, so it would be nice if `thor docs:generate ...` understood identical notation: `thor docs:generate ruby@2.7`, not only `thor docs:generate ruby --version 2.7`.  (This took me awhile to discover, after it wasn't "guessable".)

Looking at [docs:generate](https://github.com/freeCodeCamp/devdocs/blob/main/lib/tasks/docs.thor#L75-L114), teaching `docs:generate` about `@` version notation is perhaps straight-forward: `name, version = name.split('@')`.

Teaching `docs:download` about `--version` is a bit more difficult and probably unnecessary.  It would mean only allowing `--version` if one name is specified (i.e. `docs:download ruby --version=2.7`); or, pairwise matching of names and `--version=version1, version2` .. seems doubtful anyone would appreciate this CLI pattern. :)

If there is interest, perhaps I can send a PR.

richardkmichael:
Closes #1519 

@simon04 Let's try this.  I tested using a small documentation package of which there are several versions:

```
thor docs:generate sequelize    #=> Installs sequelize~6 (latest), as expected currently, given no version
thor docs:generate sequelize@5  #=> Installs sequelize~5
```

I'm less familiar with the `docs:page` command, but it should be ok.  A `nil` version as the second component of `name.split` will be handled fine (as could already happen in`find_docs()` previously) in `Docs.generate_page()` via `Docs.find()`.

Let me know if you prefer this squashed.

### DIFF
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -118,7 +118,7 @@ class DocsCLI < Thor
     puts 'Done'
   end
 
-  desc 'download (<doc> <doc@version>... | --default | --installed | --all)', 'Download documentations'
+  desc 'download (<doc> <doc@version>... | --default | --installed | --all)', 'Download documentation packages'
   option :default, type: :boolean
   option :installed, type: :boolean
   option :all, type: :boolean
@@ -141,7 +141,7 @@ class DocsCLI < Thor
     handle_doc_not_found_error(error)
   end
 
-  desc 'package <doc> <doc@version>...', 'Package documentations'
+  desc 'package <doc> <doc@version>...', 'Create documentation packages'
   def package(*names)
     require 'unix_utils'
     docs = find_docs(names)
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -65,8 +65,7 @@ class DocsCLI < Thor
     handle_doc_not_found_error(error)
   end
 
-  desc 'generate <doc> [--version] [--verbose] [--debug] [--force] [--package]', 'Generate a documentation'
-  option :version, type: :string
+  desc 'generate (<doc> | <doc@version>) [--verbose] [--debug] [--force] [--package]', 'Generate a documentation'
   option :all, type: :boolean
   option :verbose, type: :boolean
   option :debug, type: :boolean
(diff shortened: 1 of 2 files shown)


lib/tasks/docs.thor
```diff
@@ -65,8 +65,7 @@ class DocsCLI < Thor
     handle_doc_not_found_error(error)
   end
 
-  desc 'generate <doc> [--version] [--verbose] [--debug] [--force] [--package]', 'Generate a documentation'
-  option :version, type: :string
+  desc 'generate <doc> <doc@version>... [--verbose] [--debug] [--force] [--package]', 'Generate a documentation'
```
simon04, line 68:
```suggestion
  desc 'generate (<doc> <doc@version>) [--verbose] [--debug] [--force] [--package]', 'Generate a documentation'
```

As running generate on two docs does not work:
```
bundle exec thor docs:generate --force --package vue@2 vue@3
ERROR: "thor generate" was called with arguments ["vue@2", "vue@3"]
Usage: "thor docs:generate <doc> <doc@version>... [--verbose] [--debug] [--force] [--package]"
```

simon04 (changes requested):
Very nice, thank you. Please resolve the two issues I've found...

Please also update CONTRIBUTING, section "Updating existing documentations": merge step 7 into step 5?

lib/tasks/docs.thor
richardkmichael: Sorry I missed this! I must have glanced at `download` usage and copied that.

I would write a required single value with optional syntax as: `(<doc> | <doc@version>)`; using `|` as "or", consistent with the doc sets in the description of `docs:download`.  Space separated with parens still looks, to me, like more than one "docs" could be specified.

Let me know if you'd prefer your suggestion.

You mentioned two issues-- what is the other?

richardkmichael: > Please also update CONTRIBUTING, section "Updating existing documentations": merge step 7 into step 5?

I hadn't noticed this file, sorry--  now done.  Hopefully with the new version notation, how to repeat in step 7 is evident.

richardkmichael:
You're welcome, happy to help.  I'm often tethered to my phone, so local devdocs is huge data-saver for me, so thank *you* too! :-)

PR #1541, opened 2021-05-12
Dart docs latest version

jagomf:
**UPDATED**

Updated latest 2.x stable version.

---

### DIFF
--- a/lib/docs/scrapers/dart.rb
+++ b/lib/docs/scrapers/dart.rb
@@ -22,12 +22,12 @@ module Docs
 
     version '2' do
       self.release = '2.12.4'
-      self.base_url = "https://api.dart.dev/stable/#{release}/index.html"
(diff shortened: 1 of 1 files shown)


MasterEnoc:
Hi @jagomf, Dart is a FileScraper documentation, It should not make any http request for documentation. I tried to scrape the documentation without the `index.html` path and I did not get any 404 request.

jagomf:
>  I did not get any 404 request.

Did you make sure you could scrape any 2.x version other than latest stable, or any 1.x version as well?

simon04:
I could update Dart to 2.13.0 https://github.com/freeCodeCamp/devdocs/commit/314010b205fa5aaaa91cfb9803f6a62306605622 without this patch without problems.

jagomf:
Ok, I guess I had missed something.

Left only the latest docs version.

Issue #763, opened 2018-02-28
Add support to include links to the source code if available


genotrance:
Using the Nim documentation as an example, check out [this](https://nim-lang.org/docs/threads.html) link.

You can see that every method, variable or constant has a link directly to the source code in github.com. This makes it very easy to look at the actual implementation.

Other random languages I just checked. These also have such links handy. 

[Crystal](https://crystal-lang.org/api/0.24.1/Tuple.html)
[Elixir](https://hexdocs.pm/elixir/Atom.html)
[Rust](https://doc.rust-lang.org/std/primitive.char.html)

Having access to the implementation by link in devdocs will make it super useful. Without it, I end up using the language docs even though devdocs is much cooler.

As an aside, I see you have a link to the original doc page that was scraped at the bottom so in theory I could click on that, find the function again and then click there. But most docs also have a permalink to each entry. It would be beneficial to include this permalink within devdocs as well so you could jump to the official documentation from right there.

Ideally, we have both these links handy within devdocs so there's no reason to go to the official docs.

feliperaul:
100x this. I love devdocs, it's always a fixed tab in my Chrome, but the lack of the link to the actual source code in Github makes me use the Rails documentation instead of devdocs.

MasterEnoc:
@genotrance, all the documentation you mentioned have links to the source code, check them out and let me know if changes are needed.

@feliperaul Rails will have the links soon.

PR #1546, opened 2021-05-24
Elixir: include source link in section headers

breunigs:
This implements #763 just for Elixir without solving the general
issue. The style used is similar to the one of the Go docs.

2021-05-24-103603_2098x834_scrot


breunigs:
merged into https://github.com/freeCodeCamp/devdocs/pull/1546

PR #1549, opened 2021-05-26
Add documentation for WebExtensions

Cimbali:
WebExtensions from https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions

NB.:
- for lighter scraping I ran the [MDN content](https://github.com/mdn/content) locally with `yarn ; yarn start`.
- According to [the MDN copyright page](https://developer.mozilla.org/en-US/docs/MDN/About#copyrights_and_licenses), the source page should be cited with the attribution.
- There is no official icon for webextensions

### DIFF
--- a/lib/docs/filters/web_extensions/clean_html.rb
+++ b/lib/docs/filters/web_extensions/clean_html.rb
@@ -1,0 +1,14 @@
+module Docs
+  class WebExtensions
+    class CleanHtmlFilter < Filter
+      def call
+
+        # Remove all the cruft.
+        content = at_css('main#content')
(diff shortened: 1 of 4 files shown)


simon04:
Since these docs are part of MDN, please use `lib/docs/scrapers/mdn/mdn.rb` as a base class for the scraper, see `lib/docs/scrapers/mdn/svg.rb` for an example.

Cimbali:
Ah I wasn’t aware of the MDN scraper. It’s done now @simon04

simon04 (approved):
Great, thank you!

PR #1552, opened 2021-05-27
LaTeX / jq improvements

Cimbali:
These are some slight improvements 
- on the LaTeX scraper, to reduce code surrounding `<pre>` tags
- on the jq scraper, to style the examples as code − which they weren’t so far − and change the collapsing links / html mechanisms with a proper heading.
  
  The now look like:
  
  image
  image
  
  Instead of (previously):
  image
  image


simon04 (approved):
Great, thank you!

PR #1555, opened 2021-06-01
Add get_latest_version to latex

MasterEnoc:
Add get_latest_version to the LaTeX documentation. I used the date in the header of https://latexref.xyz/dev/, this is not the version of latex, but It would be helpful to know at least when the page is modified.


KenG-pro:
Thanks for this. Much needed

KenG-pro (commented):
Thanks, its a great thing

simon04 (approved):

PR #1559, opened 2021-06-07
Update Docker documentation (20.10.6)

MasterEnoc:

Looks like old versions of the docker's documentation is not available anymore, for that reason I removed old versions from the scraper.


simon04:
> Looks like old versions of the docker's documentation is not available anymore, for that reason I removed old versions from the scraper.

Since we've scraped the old docs in the past, we can still offer them on devdocs in their archived form. But I agree, we can remove the old filters and scraper options for the no longer working versions.  Basically condensing the old versions to

```ruby
    version '19' do
      self.release = '19.03'
      self.base_url = "https://docs.docker.com/"
    end
```

MasterEnoc:
Should be fine now.

simon04 (approved):
Thank you!

Issue #1553, opened 2021-05-31
Cannot add space in search box text


vixson:
# Bug report
Seach box space not working properly

## OS information
Windows 10

## Steps to reproduce
- Click on the devdoc search box
- Type a text
- Tap on your spacebar key (To give a space)
- Type another text
You'll realize that it does not give space at all. Instead, it just scrolls the page down.

## More resources
![nospace](https://user-images.githubusercontent.com/63640577/120210044-401baa00-c227-11eb-86d1-1b42833cdae0.png)
I was literally tapping the spacebar but it keeps scrolling down the current document that was opened.

## Possible fix
- Add an autocomplete feature, maybe.

> Although I discovered a hackaround last week 20th May 2021.
What I do was to just tap on my esc key then Voilà, I would be able to give space in the search box again.
But last night 30th May 2021.
If I should just tap on the esc key it just clears the entire text 😞

shashank1207:
Hi @vixson, I think it's happening because of this condition
![Screenshot from 2021-06-01 17-02-21](https://user-images.githubusercontent.com/73484229/120317129-e2a55d00-c2fb-11eb-98f5-20b088059fa1.png)
The spacebar will only work if the duration between last Keypress and spacebar is less than 500ms.

vixson:
> Hi @vixson, I think it's happening because of this condition
> ![Screenshot from 2021-06-01 17-02-21](https://user-images.githubusercontent.com/73484229/120317129-e2a55d00-c2fb-11eb-98f5-20b088059fa1.png)
> The spacebar will only work if the duration between last Keypress and spacebar is less than 500ms.

Ok @shashank1207 
But can at least, let the Keypress timeout be 1s - 3s.
😕 That half a second is too quick for us humans.
Thanks

shashank1207:
Hi. can I work on this issue? @ojeytonwilliams

ojeytonwilliams:
Absolutely, @shashank1207, a PR would be very welcome!

MasterEnoc:
@vixson this should be fixed now. You can now go to [devdocs settings](https://devdocs.io/settings) and change the time needed to scroll with spacebar that suits you better.
![image](https://user-images.githubusercontent.com/52229226/122627001-49e83e80-d06a-11eb-8220-9638b2509690.png)

vixson:
Thanks @MasterEnoc this much better now.

PR #1570, opened 2021-06-10
Changes the timelimit between spacebar and last keypress

shashank1207:
Related to the issue [#1553](https://github.com/freeCodeCamp/devdocs/issues/1553). Changed the allowed timelimit between spacebar and last keypress from 500ms to 3000ms.

### DIFF
--- a/assets/javascripts/app/shortcuts.coffee
+++ b/assets/javascripts/app/shortcuts.coffee
@@ -59,7 +59,7 @@ class app.Shortcuts
         @trigger 'escape'
         false
       when 32
-        if event.target.type is 'search' and (not @lastKeypress or @lastKeypress < Date.now() - 500)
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
LGTM :+1: 

Thanks @shashank1207, I've checked it out locally and this works nicely.  This this is a global change (i.e. not just one doc) I'll wait for one more review, just in case I'm missing something obvious.

MasterEnoc:
I think it would be better to add an option in the options page of devdocs to change individually the keypress timeout for the spacebar or change the current shortcut of the spacebar, but it could take some time. Meanwhile, we can merge this PR and see if other people think three seconds are too much for wait or not.

KenG-pro (approved):

shashank1207:
@MasterEnoc I am working on it. I will create pull request as soon as it completes.

PR #1573, opened 2021-06-14
Scroll by space

shashank1207:
Related to the issue [#1553](https://github.com/freeCodeCamp/devdocs/issues/1553#issue-907541390) and as suggested by @MasterEnoc in [this comment](https://github.com/freeCodeCamp/devdocs/pull/1570#issuecomment-859984211). 

Added a checkbox under scrolling section in preferences page and changed the allowed timelimit between spacebar and last keypress to 1000ms if the option is enabled.

### DIFF
--- a/assets/javascripts/app/settings.coffee
+++ b/assets/javascripts/app/settings.coffee
@@ -34,7 +34,7 @@ class app.Settings
     schema: 1
     analyticsConsent: false
     theme: 'auto'
-    spaceScroll: 0
+    spaceScroll: 1
 
   constructor: ->
     @store = new CookiesStore
--- a/assets/javascripts/app/settings.coffee
+++ b/assets/javascripts/app/settings.coffee
@@ -13,6 +13,7 @@ class app.Settings
     'size'
     'tips'
     'autoInstall'
+    'spaceScroll'
   ]
 
   INTERNAL_KEYS = [
@@ -33,6 +34,7 @@ class app.Settings
     schema: 1
     analyticsConsent: false
     theme: 'auto'
+    spaceScroll: 0
 
   constructor: ->
     @store = new CookiesStore
--- a/assets/javascripts/app/settings.coffee
+++ b/assets/javascripts/app/settings.coffee
@@ -14,6 +14,7 @@ class app.Settings
     'tips'
     'autoInstall'
     'spaceScroll'
+    'spaceTimeout'
   ]
 
   INTERNAL_KEYS = [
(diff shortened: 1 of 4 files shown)


assets/javascripts/app/shortcuts.coffee
```diff
@@ -59,7 +62,7 @@ class app.Shortcuts
         @trigger 'escape'
         false
       when 32
-        if event.target.type is 'search' and (not @lastKeypress or @lastKeypress < Date.now() - 500)
+        if event.target.type is 'search' and @spaceScroll() and (not @lastKeypress or @lastKeypress < Date.now() - 1000)
```
ojeytonwilliams, line 65:
```suggestion
        if event.target.type is 'search' and @spaceScroll() and (not @lastKeypress or @lastKeypress < Date.now() - 500)
```
Since space scrolling can be toggled, I think we should stick with the original delay. Existing users are used to it and a 1 second delay feels surprisingly long when you have to wait.

ojeytonwilliams (changes requested):
Other than a suggestion about the delay, this looks good to me.  Thanks @shashank1207!

ojeytonwilliams (approved):
LGTM :+1: 

@MasterEnoc is this what you had in mind?

MasterEnoc:
This does not solve the issue at all, I thought about a field (textbox) that lets user change the timeout needed. Check at my patch below:

[patch.zip](https://github.com/freeCodeCamp/devdocs/files/6667173/patch.zip)

Let me know what you think @ojeytonwilliams @shashank1207.

I think this PR plus the patch will solve the issue.

assets/javascripts/app/settings.coffee
```diff
@@ -33,6 +34,7 @@ class app.Settings
     schema: 1
     analyticsConsent: false
     theme: 'auto'
+    spaceScroll: 0
```
MasterEnoc, line 37: This should be `1` since the normal behaviour is use the space for scrolling.

MasterEnoc (changes requested):

KenG-pro (approved):

ojeytonwilliams:
The patch seems good to me.  The only thing I would suggest is a change to the wording.  How about "Delay until you can scroll by pressing space" instead of "Timeout needed to use spacebar for scrolling during search"? This phrasing should make it obvious that you have to wait before pressing space and how long you have to wait.

shashank1207:
Hi @MasterEnoc, the patch file looks good to me. I've applied the patch file and it works nicely. Also, it needed some changes in the keys in settings.coffee. I've also changed the wording as suggested by @ojeytonwilliams.

ojeytonwilliams:
Hey @shashank1207 could you push those changes?  It sounds like we're all on the same page, though.

MasterEnoc:
Everything looks good to me. I'm going to merge this PR and close #1570. Thank you @shashank1207!

PR #1574, opened 2021-06-17
docs: add unmaintained section

ojeytonwilliams:
This is for projects that are still interesting, but are somewhat neglected.

As discussed here: https://github.com/freeCodeCamp/devdocs/pull/1572#issuecomment-862861426

@MasterEnoc I'm fine with deleting it, if you'd rather, but I thought it would be nice to keep older projects around for a while.


EnginnerDog:
This pull request is magnificent. HALLELUJAH!

PR #1579, opened 2021-06-20
fix: allow 0.1 interval steps for spaceTimeout

ojeytonwilliams:
If the step is not specified, the default validation requires that the values are integers. Since the default value is 0.5, this prevented the settings from being changed.

Now the step is 0.1 which should give enough control without clashing with the default

Closes: https://github.com/freeCodeCamp/devdocs/issues/1578


MasterEnoc:
LGTM!

PR #1581, opened 2021-06-24
Avoid showing analitics notification more than once

MasterEnoc:
I usually do not accept cookies in any website and I didn't accept cookies in Devdocs. Devdocs can be really intrusive to show a notifications telling me to accept analytics each time I click a page, this is a silly behavior so I change it to only show one time the analytics notification during each Devdocs session.

I also notice that the cookies and the analytics notifications didn't hide automatically thus I added two seconds to hide them automatically.

Look at the following image showing the notifications generated by devdocs when I did not accept the cookies:
image

### DIFF
--- a/assets/javascripts/app/app.coffee
+++ b/assets/javascripts/app/app.coffee
@@ -11,6 +11,7 @@
     try @initErrorTracking() catch
     return unless @browserCheck()
 
+    @showAnaliticsOne = false
     @el = $('._app')
     @localStorage = new LocalStorageStore
     @serviceWorker = new app.ServiceWorker if app.ServiceWorker.isEnabled()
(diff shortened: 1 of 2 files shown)


assets/javascripts/lib/page.coffee
```diff
@@ -210,7 +212,7 @@ track = ->
     # Only ask for consent once per browser session
     Cookies.set('analyticsConsentAsked', '1')
```
simon04, line 213: How does your change relate to those two lines? You to tackle the same problem.

assets/javascripts/app/app.coffee
```diff
@@ -11,6 +11,7 @@
     try @initErrorTracking() catch
     return unless @browserCheck()
 
+    @showAnaliticsOne = false
```
simon04, line 14: Should this be `analiticsAlreadyShown`?

assets/javascripts/lib/page.coffee
MasterEnoc: This line avoids show me more than once the analitics consent when cookies are enabled, In my case I do not accept neither the devdocs cookie nor the analitics, thus this line does not fix my problem.

assets/javascripts/app/app.coffee
MasterEnoc: Fixed, should be fine now.

Issue #1530, opened 2021-04-27
Enable All


Shakespeare17:
# Feature request
Add an Enable All button to the offline Documentation enable/disable screen

ThePotatoChronicler:
This would be a really nice feature, but a search all option would also be useful

crazyaccess:
I would like this one to be there too, 
Rightnow whenever i do devdocs on a new system its generelly me running on the console : 
```
Array.from(document.querySelectorAll("._disabled-list span._list-enable")).map(el => {
    el.click();
    console.log(el);
})
```
(a scraper friend made it for me)

ojeytonwilliams:
Sure, this seems reasonable to me.  PRs are very welcome.

chahakshahcs5:
`const allDocs = Array.from(document.getElementsByClassName("_list-checkbox")).map(el => el.name).join("/");`
copy the value of data create devdcos.json file and insert this lines

`{
    "analyticsConsentAsked": "1",
    "docs": "[data value]"
}`

You can import this file in preference setting.

weilonger:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

chahakshahcs5:
`Array.from(document.querySelectorAll("span._list-enable")).forEach(el => el.click())`

PR #1591, opened 2021-07-27
Allow enable all documentations feature

MasterEnoc:
Adds feature in #1530.

The 'enable all' buttons can be changed to the bottom or maybe something different can be done with it.


j-f1:
Does having all the docs enabled result in performance issues?

boehs:
> Does having all the docs enabled result in performance issues?

Yes

Issue #1577, opened 2021-06-19
Rename DOM to Web API


U-C-S:
DOM is a part of Browser API / Web API.
But here, The name is used to contain all the Browser APIs and including itself, which can be quite confusing.

The Actual MDN source uses the name - "[Web API](https://developer.mozilla.org/en-US/docs/Web/API)" and [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) is just the sub-section of it

In short, Why not rename DOM to Browser API / Web API ??

KenG-pro:
Looks great

PR #1646, opened 2021-10-27
fix: Reflect title of web api over dom per issue #1577

JackHowa:
- Following recommendation in issue #1577 
- Following documentation for options title and root title: 

```md
  - `:title` [String or Boolean or Proc]
    Unless the value is `false`, adds a title to every page.
    If the value is `nil`, the title is the name of the page as determined by the [`Entries`](./filter-reference.md#entriesfilter) filter. Otherwise the title is the String or the value returned by the Proc (called for each page, with the filter instance as argument). If the Proc returns `nil` or `false`, no title is added.
  - `:root_title` [String or Boolean]
    Overrides the `:title` option for the root page only.
```

cc @U-C-S

would appreciate it if this was admitted with hacktoberfest:

A pull request is considered approved once it has an overall approving review from maintainers, or has been merged by maintainers, or has been given the 'hacktoberfest-accepted' label.*

### DIFF
--- a/lib/docs/scrapers/mdn/dom.rb
+++ b/lib/docs/scrapers/mdn/dom.rb
@@ -7,7 +7,8 @@ module Docs
 
     html_filters.push 'dom/clean_html', 'dom/entries'
 
-    options[:root_title] = 'DOM'
+    options[:title] = 'Web API'
+    options[:root_title] = 'Web API'
 
   end
 end


JackHowa:
closing this due to inactivity cc @u-c-s

U-C-S:
@JackHowa let it stay open. They (maintainers) will look into this someday.

JackHowa:
looks like this was done! @U-C-S

PR #1670, opened 2021-12-05
Remove FastClick library

simon04:
From https://github.com/ftlabs/fastclick/blob/main/README.md
> Note: As of late 2015 most mobile browsers - notably Chrome and Safari - no longer have a 300ms touch delay, so fastclick offers no benefit on newer browsers, and risks introducing bugs into your application. Consider carefully whether you really need to use it.


MasterEnoc:
Looks good to me, since modern browsers do not need fastclick and devdocs recommends modern
browsers to be run (at least without issues).

KenG-pro:
Look ok

PR #1686, opened 2022-01-07
Respect Do Not Track and Global Privacy Control

### LINKED ISSUE #1685: Devdocs should respect the Do Not Track header
Having enabled [Enhanced Tracking Protection in Firefox](https://support.mozilla.org/en-US/kb/enhanced-tracking-protection-firefox-desktop?as=u&utm_source=inproduct), I do expect Devdocs not to ask me whether I'm okay with tracking cookies.

Firefox sends the [DNT=1 header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT). According to https://github.com/mdn/content/pull/4960 this header is considered deprecated as of 2021, but I'm not aware of an alternative. 😕

simon04:
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack
- https://globalprivacycontrol.org/ via `navigator.globalPrivacyControl`

Fixes #1685

### DIFF
--- a/assets/javascripts/lib/page.coffee
+++ b/assets/javascripts/lib/page.coffee
@@ -200,6 +200,8 @@ page.track = (fn) ->
 
 track = ->
   return unless app.config.env == 'production'
+  return if navigator.doNotTrack == '1'
+  return if navigator.globalPrivacyControl
 
   consentGiven = Cookies.get('analyticsConsent')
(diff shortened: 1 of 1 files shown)


ojeytonwilliams (approved):
LGTM :+1:

PR #1694, opened 2022-01-15
Improve point_cloud_library clean_html

ardiya:
This commit tries to improve the clean_html with 2 updates:

**1. Improve table appearance by removing unnecessary gap between table**

Before:
image

After:
image

**2. Improve code appearance**

Before:

image

After:

image

cc: @kunaltyagi <- PCL maintainer who requested this PR


xjkdev:
Hi! Because PCL's documents are also generated by doxygen, there is a same issue that the space indents in code fragments were removed. And I have solved this problem by writing filter codes to`parse` function in `UrlScrapper` instead.

And I believe it can be better if you do the same. please check #1717 and lib/docs/scrapers/eigen3.rb .

PR #1718, opened 2022-03-14
Add docs for process_response? and parse.

xjkdev:
Add description of `process_response?` method and `parse` method in scrappers' document.


simon04 (approved):
Thank you!

PR #1738, opened 2022-04-24
Add nix and nixpkgs

thatsmydoing:
This only adds a subset of nix and nixpkgs functions. Notably derivation, stdenv and language-specific functions are omitted for now as their formats differ significantly and will require more cleanup.

- [X] Tested the scraper on a local copy of DevDocs
- [X] Ensured that the docs are styled similarly to other docs on DevDocs
- [X] Added these files to the <code>public/icons/*your_scraper_name*/</code> directory:
  - [X] `16.png`: a 16×16 pixel icon for the doc
  - [X] `16@2x.png`: a 32×32 pixel icon for the doc
  - [X] `SOURCE`: A text file containing the URL to the page the image can be found on or the URL of the original image itself


simon04 (approved):
Thank you!

wmertens:
@thatsmydoing you could add the stdenv etc documentation as guides? Thank you for doing this in any case, very useful :-)

PR #1747, opened 2022-05-05
fix: clip hover state breaking column layout

### LINKED ISSUE #1739: "Copy to Clipboard" icon on :hover state breaks column layout
# Bug report

The "copy to clipboard" icon's :hover state breaks the table layout when the row's text occupies nearly the entire width of the column. See gifs and cause below.

Fix is included below; if you'd like me to open a PR with it let me know.

## OS information

Windows 10, using Google Chrome Version 100.0.4896.127 (Official Build) (64-bit). Also had the same issue on iPad using the Google search app (ver <= 130.0.X) on iOS. Browser font zoom level set at default 100%.

## Steps to reproduce

Hover over the `pre` region of a `t-dcl` row so the the "copy to clipboard" icon appears. The bug is most noticeable when the difference between the column width, and the right-most edge of that row's last `token` span is less than 20px (the size of the browser's vertical scrollbar's width), forcing an overflow when the icon display is set to visible.

For rows containing a single line of text, an inline scrollbar can be seen to the right of the clipboard icon (due to the size of the `_pre-clip` button being taller than that of the row's text height).

If the inline scrollbar forces that single row of text to overflow into a second line, dependent on where the line break occurs, it can often remove itself briefly before coming back again, causing the entire table's layout to rapidly jitter.

## More resources

The bug:

chrome_PGWDqEURnI

With padding/positioning fix:

chrome_qNr9HrK3CD

## Possible fix

The reason for layout changes on hover is from the width of the scrollbar being added to the `pre` element. So either allow the button to overflow without affecting the containing element's layout, or adjust the icon's size (including padding) so there is no overflow.

Reducing vertical space used by the clipboard icon's button element to be <= the parent `pre`'s height, with a minor positioning tweak to compensate, fixes the problem:

https://github.com/freeCodeCamp/devdocs/blob/28a98a899473cc3a5b490477c15d8751410efb84/assets/stylesheets/components/_content.scss#L401

```css
pre:hover>._pre-clip {
    display: block;
    top: 0.1875rem; /* adjust for removed padding */
    padding: 0; /* override default padding */
}
._pre-clip {
    display: none;
    position: absolute;
    top: 0;
    right: 0;
    opacity: .5;
    padding: 0.375rem; /* causing clipboard icon bug */
    cursor: pointer;
}
```

Changing the `overflow` property to "visible" or "inherit" for the `pre,._jq .manual-example table td,._redis>.example,._scala .related-types` style also fixes it.

```css
pre, ._jq .manual-example table td, ._redis>.example, ._scala .related-types {
    position: relative;
    margin: 1.5em 0;
    padding: 0.375rem 0.625rem;
    line-height: 1.5;
    /* overflow: auto; */
    overflow: inherit; /* or visible */
}
```

StelioKontosXBL:
Minor positioning/padding changes to the "copy to clipboard" icon's style rule to prevent visible layout changes from occurring while in the :hover state. fixes #1739

### DIFF
--- a/assets/stylesheets/components/_content.scss
+++ b/assets/stylesheets/components/_content.scss
@@ -398,7 +398,12 @@
   padding: .375rem;
   cursor: pointer;
 
-  pre:hover > & { display: block; }
+  pre:hover > & {
+    display: block;
+    top: 0.1875rem;
+    padding: 0;
+  }
+
   &:hover { opacity: 1; }
 
   > svg {


simon04 (approved):
Thank you!

PR #1746, opened 2022-05-04
Add Elixir v1.13.4

opie4624:
If you're updating existing documentation to its latest version, please ensure that you have:

- [x] Updated the versions and releases in the scraper file
- [x] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches its data in `self.attribution`
- [x] Ensured the icons and the `SOURCE` file in <code>public/icons/*your_scraper_name*/</code> are up-to-date if the documentation has a custom icon
- [x] Ensured `self.links` contains up-to-date urls if `self.links` is defined
- [ ] Tested the changes locally to ensure:
  - The scraper still works without errors
  - The scraped documentation still looks consistent with the rest of DevDocs
  - The categorization of entries is still good

PR #1758, opened 2022-05-27
Fix all Elixir 1.13 names detected as "Settings"

breunigs:
A partial fix was already included in the initial patch in #1746,
but it was only applied to a subset of titles.

I tested the new scraper for both Elixir 1.13 and the previous 1.12 and it still looks good


simon04 (approved):
Thank you!

PR #1809, opened 2022-09-02
Fix MDN iframe height

FireIsGood:
I found that the height of the [_MDN.scss](https://github.com/freeCodeCamp/devdocs/blob/main/assets/stylesheets/pages/_mdn.scss) file was not rendering the iframe .content class correctly and using its default 200px height as compared to the original.

This commit tries to fix the iframe element have a higher default height to make the scrollbar go away and copy the original website's css to make the iframe become taller on mobile and thin screens.

## Before:
old full
old thin

## After:
new full
new thin

This solution generally improves the readability but isn't a perfect fix since the media query can't calculate for the sidebar width as far as I am aware. The iframe overflow is tuned for a 1920x1080 screen but will be off by a few pixels depending on the screen size.

### DIFF
--- a/assets/stylesheets/pages/_mdn.scss
+++ b/assets/stylesheets/pages/_mdn.scss
@@ -132,7 +132,50 @@
     dd { margin: 0; }
   }
 
-  iframe.interactive {
-    width: 100%;
+  .interactive {
+      width: 100%;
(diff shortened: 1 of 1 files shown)


simon04 (approved):
Great, thank you!

PR #1817, opened 2022-09-19
fix: replace calls to `File.exists?` with `File.exist?`

andyholmes:
This prints warnings about being deprecated, so replace it with the call that the deprecated function wraps.

See: https://github.com/ruby/ruby/blob/ruby_2_7/file.c#L1752-L1781


simon04 (approved):
Thank you!

PR #1839, opened 2022-10-11
Fix Erlang lib/ entries and func Types sec

### LINKED ISSUE #795: Erlang documentation is not correctly indexed
As of commit d48396d, the Erlang documentation is not correctly indexed and/or not correctly searched:

There are two problems, since they may be related, I put both of them here:

1. Search with `lists:sum` should show documentation about the function `lists:sum`, but empty results are returned instead. (not all are badly indexed, for example, search `gen_server:call` works as expected.

image

2. Some functions of the `lists` module is missing in the export list. See the below two screenshots.

Wrong behaviour (only one export is shown)
image

Expected behaviour (many exports are shown)
image

axelf4:
This pull request

* Makes lib/ entries get indexed correctly;
* Stops the types associated with functions from being filtered away, and orders their "Types" heading under the function heading; and
* Cleans out onmouseover/onmouseout attributes.

Closes #795

### DIFF
--- a/assets/stylesheets/pages/_erlang.scss
+++ b/assets/stylesheets/pages/_erlang.scss
@@ -6,4 +6,10 @@
   .note { @extend %note; }
   .warning { @extend %note, %note-red; }
   .note .label, .warning .label { font-weight: var(--boldFontWeight); }
+
+  .since {
+	  color: var(--textColorLight);
+	  font-weight: normal;
+	  font-size: small;
+  }
 }
--- a/lib/docs/filters/erlang/clean_html.rb
+++ b/lib/docs/filters/erlang/clean_html.rb
@@ -30,34 +30,47 @@ module Docs
 
         # others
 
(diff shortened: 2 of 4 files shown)


simon04:
Thanks a lot for this PR. The results look very promising. What's left to be done?

Just to make sure, here's how to provide the source docs: https://github.com/freeCodeCamp/devdocs/blob/main/docs/file-scrapers.md#erlang

axelf4:
> What's left to be done?

I had only been concerned with OTP 25 initially. I have now to the best of my ability gone through older versions to ensure their indexing is also correct, and that there are no regressions.

simon04 (approved):
Thank you!

simon04:
Great! I've updated Erlang v25 and v24.

Stratus3D:
This is awesome! 🔥 Thank you @axelf4 !

PR #1849, opened 2022-10-23
Update octave scraper to 7.2.0

foo-dogsquared:

### DIFF
--- a/assets/javascripts/templates/pages/about_tmpl.coffee
+++ b/assets/javascripts/templates/pages/about_tmpl.coffee
@@ -645,7 +645,7 @@ credits = [
     'Octave',
     '1996–2022 The Octave Project Developers',
     'Octave',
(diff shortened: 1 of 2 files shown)


foo-dogsquared:
> The second day in a row I cannot access docs.octave.org
> 
> ```
> > bundle exec thor docs:generate --force --package octave
> ERROR:                                                                                                                                                                                         
>   https://docs.octave.org/v7.2.0/index.html
>   RuntimeError: Error status code (0): Timeout was reached
>     https://docs.octave.org/v7.2.0/index.html
> ```

I currently can generate one. Though I did encounter the same issue for the past few days. Seems to be on the server side as I also cannot access the original pages from the current version from the devdocs.io instance (with `alt+o` shortcut).

simon04 (approved):
Thank you!

PR #1877, opened 2022-11-27
Update Dart (2.18.5)

ClasherKasten:

- [X] Ensured the icons and the `SOURCE` file in <code>public/icons/docs/dart/</code> are up-to-date if the documentation has a custom icon


simon04 (approved):
Thank you!

PR #1911, opened 2023-01-10
fix: download db.json in production

ojeytonwilliams:
Closes https://github.com/freeCodeCamp/devdocs/issues/1910

@simon04 there's probably a better approach where we point the app at documents.devdocs.io (since https://documents.devdocs.io/bash/db.json and the like do exist), but this was the best I could up with for now!

### DIFF
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -238,7 +238,7 @@ class DocsCLI < Thor
           dir = File.join(Docs.store_path, doc.path)
           FileUtils.mkpath(dir)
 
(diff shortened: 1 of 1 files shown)


simon04:
I @ojeytonwilliams, thank you for looking into this issue. It's a regression of #1293. It's an issue not reproducible using the local development server `bundle exec rackup`. Please let me try to address the regression...

ojeytonwilliams:
You're welcome - I thought it was worth me having a go, even though I knew you'd be on top of it!

PR #1912, opened 2023-01-10
Fix manifest db_size

### LINKED ISSUE #1910: Offline size shown as "NaN MB"
# Bug report

Today I opened devdocs.io on my browser and found many documentations have been updated. Okay, no problem, so I went to the "Offline" page and was surprised to see "NaN MB" as the size of each documentation. Okay, maybe I can uninstall all, refresh, install all, and it will be solved… Nope, even after uninstalling all and reinstalling all, it still shows "NaN MB".

## OS information

Mozilla Firefox 108.0.1 (64-bit) on Manjaro Linux.

## Steps to reproduce

Use devdocs.io on Firefox, and download offline documentation.

I have NOT tried on Chrome (yet).

I have NOT tried on a new Firefox profile.

~~I have NOT tried to empty all local storage and reload devdocs.io. I don't want to try it because I'm tired of re-selecting all the documentation every time I switch browsers, or every time the browser decides to delete devdocs local storage. Seriously, I lost count of how many times I've done that.~~ I tried the red "Reset" link at the settings page, and this issue persists.

## More resources

Screenshot

## Possible fix

I don't know how to fix it.

~~To ease the pain, it would be awesome to store the devdocs settings and list of enabled documentations somewhere else. Maybe a manual import/export, maybe syncing on some remote server, well, something. But this wish is off-topic, as it should become a different issue.~~ I just found out about #1818, and I had never noticed those Export/Import buttons before.

simon04:
Regression of #1293.

Fixes #1910.

Closes #1911.

### DIFF
--- a/lib/docs/core/doc.rb
+++ b/lib/docs/core/doc.rb
@@ -86,16 +86,6 @@ module Docs
         json
       end
 
-      def as_json_extra(store)
-        json = self.as_json
-        if options[:attribution].is_a?(String)
-          json[:attribution] = options[:attribution].strip
-        end
-        json[:db_size] = store.size(self.db_path) if store.exist?(self.db_path)
-        json[:mtime] = store.mtime(self.meta_path).to_i if store.exist?(self.meta_path)
-        json
-      end
-
       def store_page(store, id)
         index = EntryIndex.new
         pages = PageDb.new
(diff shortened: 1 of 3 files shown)


ojeytonwilliams (approved):
LGTM :+1: 

Everything works fine locally and the code makes sense. I'll merge it now so we can see if it works in prod.

PR #1913, opened 2023-01-12
Update Python (3.10.9, 3.11.1); Update attribution (it's 2023)

ClasherKasten:

- [X] Ensured the icons and the `SOURCE` file in <code>public/icons/docs/python/</code> are up-to-date if the documentation has a custom icon

## Annotation

The scraper works very well locally, but it seems that there was some problem with `devdocs.io` for Python 3.10.

### Reproduction:
```sh
mkdir docs/python~3.10
cd docs/python~3.10
curl -L https://docs.python.org/3.10/archives/python-3.10.9-docs-html.tar.bz2 | \
    tar xj --strip-components=1
cd ../..
bundle exec thor docs:generate python@3.10
bundle exec rackup
```
gives the following output when searching for `match` (it is expected to find the new introduced `match`-statement):

local output

The output matches the expectation.
In contrast the output `devdocs.io` gives:

devdocs.io output

Which doesn't show the `match`-statement at all (it's also not to be found when scroling further down or looking manually through the documentation).

I honestly don't know how that difference was possible.

### DIFF
--- a/lib/docs/scrapers/python.rb
+++ b/lib/docs/scrapers/python.rb
@@ -17,19 +17,19 @@ module Docs
       library/sunau.html)
 
     options[:attribution] = <<-HTML
-      &copy; 2001&ndash;2022 Python Software Foundation<br>
+      &copy; 2001&ndash;2023 Python Software Foundation<br>
       Licensed under the PSF License.
     HTML
 
     version '3.11' do
-      self.release = '3.11.0'
+      self.release = '3.11.1'
       self.base_url = "https://docs.python.org/#{self.version}/"
 
       html_filters.push 'python/entries_v3', 'sphinx/clean_html', 'python/clean_html'
     end
 
     version '3.10' do
-      self.release = '3.10.8'
+      self.release = '3.10.9'
       self.base_url = "https://docs.python.org/#{self.version}/"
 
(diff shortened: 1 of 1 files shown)


simon04 (approved):
Great, thank you!

simon04:
Thank you for noticing!

Err, something seems to be terribly wrong with https://devdocs.io/python~3.10/ since it says:

>  Welcome! This is the official documentation for Python 3.9.14. 

Also the 3.11 docs seem to have gotten some updates:

```
> bundle exec thor docs:download python
(1/1) Python 3.11 OK
Done

> bundle exec thor docs:generate --force --package python
Entries: (1230 KB, +1 KB)
  Updated: 11224
  Created: 7
    + (?( (in regular expressions)
    + Applications (struct)
    + as (match statement)
    + c.PyFrame_Check()
    + Distribution Discovery (Using importlib.metadata)
    + Download Python 3.11.1 Documentation
    + sqlite3.Cursor.row_factory
  Deleted: 1
    - Download Python 3.11.0 Documentation
Types:
  Updated: 48
  Created: 0
  Deleted: 0
Files: (17.1 MB, +0.1 MB)
  Updated: 448
  Created: 0
  Deleted: 0
Done
```

PR #1926, opened 2023-02-11
Update Babel documentation

ClasherKasten:

Notes:
- Updated all Babel versions that needed to be updated

Question:
- Why both maintaining Babel `7.x` and `6.x` when they both fetch from the same resource and provide exactly the same informations? Wouldn't it be easier to drop `6.x` then?

### DIFF
--- a/lib/docs/scrapers/babel.rb
+++ b/lib/docs/scrapers/babel.rb
@@ -30,11 +30,11 @@ module Docs
     HTML
 
     version '7' do
-      self.release = '7.16.4'
+      self.release = '7.20.15'
     end
 
     version '6' do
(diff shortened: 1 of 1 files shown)


simon04:
Scraper needs more than a version string update

```
> bundle exec thor docs:generate --force --package babel --debug
Queue:   babeljs.io/docs/en/                                                                                                                                                                                                                                    
Ignore:  babeljs.io/docs/en/                                                                                                                                                                                                                               [200]
[#####################################################################################################################################################################################################################] [1/1] [100.00%] [00:00] [00:00] [2.53/s]Failed!
```

PR #1932, opened 2023-02-22
fix: 🐛 babel document duplicate and do not display

ChandlerVer5:

1. Fixed these errors in the image 
  These errors are caused, when scrape an internal link, sometimes it contains `/en/`, but it duplicated 
  image

2. Fixed some pages dont display  in the left sidebar
  image


simon04 (approved):
Thank you!

PR #1934, opened 2023-03-01
fix(lib): use semver for minor versioning

ShaunSHamilton:

- [ ] Tested the scraper on a local copy of DevDocs
- [ ] Ensured that the docs are styled similarly to other docs on DevDocs

- [ ] Added these files to the <code>public/icons/*your_scraper_name*/</code> directory:
  - [ ] `16.png`: a 16×16 pixel icon for the doc
  - [ ] `16@2x.png`: a 32×32 pixel icon for the doc
  - [ ] `SOURCE`: A text file containing the URL to the page the image can be found on or the URL of the original image itself

- [ ] Ensured the license is up-to-date and that the documentation's entry in the array in `about_tmpl.coffee` matches its data in `self.attribution`

---

**Open for Discussion**

- [Original commit making change](https://github.com/freeCodeCamp/devdocs/commit/200e39ce9042fd4e254208837644486c4c2d89a6)
- `1.X` updates currently count for MAJOR changes
- This goes against https://semver.org/

Originally, I noticed this because in the monthly version report, Rust going from `1.65` to `1.67` was marked under the MAJOR updates. There are many other examples

**This PR**

- Mark `1.X` updates as MINOR

### DIFF
--- a/lib/docs/core/doc.rb
+++ b/lib/docs/core/doc.rb
@@ -211,7 +211,6 @@ module Docs
         break if i >= scraper_parts.length or i >= latest_parts.length
         return 'Outdated major version' if i == 0 and latest_parts[i] > scraper_parts[i]
         return 'Outdated major version' if i == 1 and latest_parts[i] > scraper_parts[i] and latest_parts[0] == 0 and scraper_parts[0] == 0
-        return 'Outdated major version' if i == 1 and latest_parts[i] > scraper_parts[i] and latest_parts[0] == 1 and scraper_parts[0] == 1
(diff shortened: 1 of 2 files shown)


simon04:
I've consciously added this differentiation in 200e39ce9042fd4e254208837644486c4c2d89a6 since various docs such as Elixir, Go, nginx, PyTorch, Rust, seem to be "stuck" in the 1.x.y scheme. Thus, I prefer to keep the current behaviour. Thanks for reaching out, though.

Issue #1977, opened 2023-05-04
`better-errors` v2.10.0 breaks development


ClasherKasten:
# Bug report
Cloning the latest version of `devdocs` and trying to spin up the local development raises the following error (An image to the error is appended at the end):
```
<internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- sassc (LoadError)
Did you mean?  sass
	from <internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /tmp/tests/devdocs/rbenv_broken/lib/ruby/gems/3.2.0/gems/better_errors-2.10.0/lib/better_errors/error_page_style.rb:1:in `<top (required)>'
	from <internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from <internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /tmp/tests/devdocs/rbenv_broken/lib/ruby/gems/3.2.0/gems/better_errors-2.10.0/lib/better_errors/error_page.rb:5:in `<top (required)>'
	from <internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from <internal:/tmp/tests/devdocs/rbenv_broken/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
	from /tmp/tests/devdocs/rbenv_broken/lib/ruby/gems/3.2.0/gems/better_errors-2.10.0/lib/better_errors.rb:8:in `<top (required)>'
	from /tmp/tests/devdocs/rbenv_broken/lib/ruby/gems/3.2.0/gems/bundler-2.4.6/lib/bundler/runtime.rb:60:in `require'
	from /tmp/tests/devdocs/rbenv_broken/lib/ruby/gems/3.2.0/gems/bundler-2.4.6/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
[...]

PR #1990, opened 2023-05-20
Update Gemfile for sassc dependency on MacOS Apple Silicon

JoshJarabek7:
Fixes sassc bundle exec rackup error on MacOS Apple Silicon.


RossBarnie:
This looks like it's trying to solve #1977 and should have been fixed by #1979.

PR #1993, opened 2023-05-24
update multiple minor versions

ClasherKasten:

(This checklist is complete for all docs updated in this PR)

---

### Notes:
- Flask/Werkzeug:
    - the link provided in `self.links[home]` is officially removed from the docs with no replacement. I didn't change it yet, but the question is if we still want to keep it?
    - the icon change was towards version 2.3.x. Still this also overrides the icon for previous versions which isn't the case on the official docs site. Is this a problem? if yes, would it be better to maintain multiple versions of the icon or just changing it back?
- Qt:
    - Would it an idea to remove all non-LTS versions (or at least the ones that are unsupported now) and thus reducing the maintenance load?


simon04 (approved):
Great, thank you!

I think the Qt scraper needs some rework, 38k pages are listed under "Qt". I've reverted this change for now to merge the remaining updates.

Screenshot 2023-06-11 at 11-22-39 Qt documentation — DevDocs

PR #2019, opened 2023-07-26
Update multiple "current version" checks

ClasherKasten:
- Single scrapers (just effecting one (max. two) scrapers):  GCC, GNU Fortran, React Bootstrap, Docker, Electron, C/C++
- Adjustment in Base Scraper (potential affect in all (though practically didn't))


simon04 (approved):
Great, thank you!

PR #2043, opened 2023-08-27
fix(fluture): prevent sorting types

jceb:
It's just a fix so that the structure of the documentation (types) remains intact.


simon04 (changes requested):
This seems to break anchor links. https://devdocs.io/fluture/index#encase works nicely, whereas http://localhost:9292/fluture/index#encase no longer works.

simon04 (approved):
Thank you!

PR #2044, opened 2023-08-29
fix: download_doc should delete tempfile

sigvei:
The download task in `docs.thor` did not delete the tarball after unpacking in `/tmp`.


simon04 (approved):
Thank you!

Issue #1997, opened 2023-05-31
React documentation is from the old website


FredrikMeyer:
# Documentatio bug

The React documentation links to the old React documentation at reactjs.org. The new website is react.dev.

MohdSaquib114:
Hey i want to work on this issue

gombosg:
I'm planning to have a go at this one as react.dev is 100% in-use now.

PR #2057, opened 2023-09-18
Fixed Reactjs Documentation bug 

17Ved:
#1997
Fixed the Documentation bug - 

The React documentation links to the old React documentation at reactjs.org. The new website is react.dev.

### DIFF
--- a/lib/docs/scrapers/react.rb
+++ b/lib/docs/scrapers/react.rb
@@ -3,7 +3,7 @@ module Docs
     self.name = 'React'
     self.type = 'simple'
     self.release = '18.2.0'
-    self.base_url = 'https://react.dev/'
+    self.base_url = 'https://react.dev/reference/react/'
     self.root_path = 'hello-world.html'
     self.links = {
       home: 'https://reactjs.org/',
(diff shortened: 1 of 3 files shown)


simon04 (changes requested):
The scraper needs to be adapted/fixed!

```
> bundle exec thor docs:generate --force --package react
Failed! (try running with --debug for more information)
```

lib/docs/scrapers/react.rb
```diff
@@ -3,7 +3,7 @@ class React < UrlScraper
     self.name = 'React'
     self.type = 'simple'
     self.release = '18.2.0'
-    self.base_url = 'https://reactjs.org/docs/'
+    self.base_url = 'https://react.dev/'
```
johnhunter, line 6: > The scraper needs to be adapted/fixed!
> 
> ```
> > bundle exec thor docs:generate --force --package react
> Failed! (try running with --debug for more information)
> ```

I don't know much about the scraper but I wonder if the `base_url` needs to point to the root of the API docs?

```suggestion
    self.base_url = 'https://react.dev/reference/react/'
```

simon04:
Please see https://github.com/freeCodeCamp/devdocs/blob/main/.github/CONTRIBUTING.md and open a new pull request once the adapted scraper is working. Thanks.

Issue #1610, opened 2021-09-17
C++ std::variant::operator= page is empty


MScieburakoADVA:
https://devdocs.io/cpp/utility/variant/operator=

Page is completely empty, but with correct footer and proper link to cppreference.com.
Other std::variant pages are fine.

Tested on online documentation, via Vivaldi and Edge.

atharva20-coder:
Hello, @MScieburakoADVA I want to know that where can I find this in the repository.

amedina153:
I am currently working on this issue, may I be assigned to this please?

thewheat:
Testing https://github.com/freeCodeCamp/devdocs/pull/2080 (which tries to fix https://github.com/freeCodeCamp/devdocs/issues/2039) the page is generated correctly, so I think that PR should fix this issue as well

<img width="1000" alt="image" src="https://github.com/freeCodeCamp/devdocs/assets/892961/20521d45-af72-40c5-b1e4-c70713d5550e">

PR #2080, opened 2023-10-10
Remove parent divs only once for each `ul` element

### LINKED ISSUE #2039: Documentation on C++ std::vector::vector shows no content
# Bug report

## OS information

Windows 11, Edge 115.0.1901.203

## Steps to reproduce

1. Go to documentation of std::vector::vector [Link](https://devdocs.io/cpp/container/vector/vector)
2. No content is shown, although the link to cppreference is correct, and cppreference has content. 

## More resources

image

image

mateusdeap:
Fixes #2039 

The issue was that, in `Docs::Cpref::CleanHtmlFilter`, we had the following:

```ruby
        css('div > ul').each do |node|
          node.parent.before(node.parent.children).remove
        end
```

This was fine for almost all cases, however, the page for [`std::vector::vector`](https://en.cppreference.com/w/cpp/container/vector/vector) had more than one `ul` element inside the same `div` (in the Complexity section).

This meant that, as we iterated over each `ul` element, because 2 or more had the same parent `div`, for the first `ul` with a sibling, we'd correctly pop all it's siblings out of the parent div, but when we came around to the second sibling `ul`, it's parent `div` was now some outer `div` on the page, and that would get removed. This would repeat itself how many sibling `ul` elements there might be on the page.

I modified the code to simply grab the parent divs, remove the duplicates and then execute the logic to pop out it's children and remove itself.

All tests passed locally, but when I went ahead to add tests to prevent this bug from reappearing, I noticed that there doesn't seem to be any tests for specific scrapers or filters, is that by choice?

### DIFF
--- a/lib/docs/filters/cppref/clean_html.rb
+++ b/lib/docs/filters/cppref/clean_html.rb
@@ -19,8 +19,9 @@ module Docs
           node.before(node.children).remove
         end
 
-        css('div > ul').each do |node|
-          node.parent.before(node.parent.children).remove
+        parents = css('div > ul').map(&:parent).uniq
+        parents.each do |parent|
+          parent.before(parent.children).remove
         end
 
         css('dl > dd:first-child:last-child > ul:first-child:last-child').each do |node|


thewheat:
This also sees to fix #1610

jasonbtiller:
This looks great. Why hasn't it been merged?

mateusdeap:
I do not know hahaha I just assumed that, eventually, the maintainer would se it.

simon04 (approved):
Thank you!

PR #2096, opened 2023-11-20
Fix the regexp for Golang generic functions

sharpevo:
Fix the regular expression of additional entries for Golang funciton signatures that are generic.

Before:
before

After:
after


simon04 (approved):
Great, thank you!

PR #2104, opened 2024-01-03
Update Ruby, Rails, and Minitest docs

larouxn:
Heavily based on the following.
- https://github.com/freeCodeCamp/devdocs/pull/1902

Sources
- https://www.ruby-lang.org/en/downloads/releases/
- https://rubyonrails.org/category/releases
- https://github.com/minitest/minitest/tags


larouxn:
👋 @simon04, perhaps you could have a look at this when you have a moment. 🙇‍♂️

simon04 (approved):
Thank you!

PR #2107, opened 2024-01-06
Upgrade ActiveSupport from 7.0.8 to 7.1.2

larouxn:
Upgrades ActiveSupport from 7.0.8 to 7.1.2 and removes deprecated `per_thread_registry` usage.

- [7.1.2 Changelog](https://github.com/rails/rails/blob/6b93fff8af32ef5e91f4ec3cfffb081d0553faf0/activesupport/CHANGELOG.md)

This upgrade and fix was inspired by https://github.com/freeCodeCamp/devdocs/pull/2073#issuecomment-1879764828. Thus, this PR is an alternative PR to the following bump PR. Ideally one or the other is merged and the other closed.
- https://github.com/freeCodeCamp/devdocs/pull/2073

Tests work fine locally. CI looks good too.

Screenshot from 2024-01-07 08-17-39

### DIFF
--- a/lib/app.rb
+++ b/lib/app.rb
@@ -56,7 +56,6 @@ class App < Sinatra::Application
 
     SpritesCLI.new.invoke(:generate, [], :disable_optimization => true)
 
-    require 'active_support/per_thread_registry'
     require 'active_support/cache'
     sprockets.cache = ActiveSupport::Cache.lookup_store :file_store, root.join('tmp', 'cache', 'assets', environment.to_s)
   end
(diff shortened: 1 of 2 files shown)


lib/app.rb
```diff
@@ -56,7 +56,6 @@ class App < Sinatra::Application
 
     SpritesCLI.new.invoke(:generate, [], :disable_optimization => true)
 
-    require 'active_support/per_thread_registry'
```
larouxn, line 59: This was added in https://github.com/freeCodeCamp/devdocs/commit/07134dd1443957e92a32ce791cdb99d80ee18d09 9 years ago to "Add missing require for active_support/cache". Might not be (probably isn't) necessary anymore.

simon04 (approved):
Great, thank you!

PR #2109, opened 2024-01-09
Upgrade `sinatra` and `siantra-contrib` from `3.1.0` to `3.2.0` + related gems

larouxn:
Upgrades `sinatra` and `siantra-contrib` from `3.1.0` to `3.2.0` in addition to bumping sub-dependencies. Extracted bump from a `bundle update` call as individual `bundle update sinatra` was not working. 

https://github.com/sinatra/sinatra/blob/fa5a1399b6b63bcca2df81db284ff64a89a2b599/CHANGELOG.md#320--2023-12-29

This bump is good to have in general but is also primarily for Ruby 3.3.0 support.

https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/


simon04 (approved):
Thank you!

PR #2127, opened 2024-01-16
Enable Ruby YJIT for performance boost 🚀

larouxn:
Follow up to the upgrade to Ruby 3.3.0, which brought with a much improved YJIT, this PR actually enables said YJIT for a supposed free performance boost of around 15%-25% latency wise.
- https://github.com/freeCodeCamp/devdocs/pull/2110

Caveat being devdocs isn't a Rails codebase nor is it a massive app but regardless, this should be a free performance win with little to no possible downsides **_assuming devdocs is not running with minimal RAM_**. For further reference, here is a blog post on the performance increase, as well as PR that made YJIT easily enable-able for Rails on Ruby 3.3+.
- https://railsatscale.com/2023-12-04-ruby-3-3-s-yjit-faster-while-using-less-memory/
- https://github.com/rails/rails/pull/49947

As for the approach, I don't know a whole lot about Dockerfiles nor setting arbitrary env vars but I followed the following two guides/posts on how to enable YJIT, as they seem legitimate, having uses fly.io before myself.
- https://mailsnag.com/blog/optimized-ruby-dockerfile/
- https://fly.io/docs/rails/cookbooks/deploy/#yjit


simon04:
Hi, the Docker files are provided only for people who want to self-host. The main instance devdocs.io is running on Heroku and is being deployed via `.github/workflows/build.yml` – do you think it's worth investigating/enabling YJIT on Heroku?

(@ojeytonwilliams, fyi)

larouxn:
> Hi, the Docker files are provided only for people who want to self-host. The main instance devdocs.io is running on Heroku and is being deployed via `.github/workflows/build.yml` – do you think it's worth investigating/enabling YJIT on Heroku?
> 
> (@ojeytonwilliams, fyi)

Ah, understood. It looks like all we need to do to enable YJIT on Heroku is set the variable, according to the [docs](https://devcenter.heroku.com/articles/ruby-support#yjit).

```console
heroku config:set RUBYOPT="--enable-yjit"
```

I suppose we can close this PR unless we want to recommend enabling YJIT by default, which seems reasonable.

larouxn:
Closing this as the change can be made at the application level re: https://github.com/freeCodeCamp/devdocs/pull/2127#issuecomment-1902809386.

PR #2141, opened 2024-01-24
Update kubernetes.rb to add versions available

daodennis:
Previously only the latest and v1.20 was available, this change adds the latest version 1.29 and the previous five versions through 1.24 which are common K8S releases in circulation.

TBD on these:
- [ ] Ensured the license is up-to-date


simon04 (approved):
Thank you!

PR #2174, opened 2024-03-08
Update README to include new Vim plugin (to browse docs without leaving Vim)

girishji:
Include the following Vim plugin in README:
https://github.com/girishji/devdocs.vim

With this plugin you can:
- browse documents within Vim
- navigate links (like in a real browser)
- install/uninstall slugs
- use VIm as a pager (TUI)

None of the other Vim/Neovim plugins have all these features.


girishji:
Please merge this PR. It has been 2 months.

PR #2221, opened 2024-05-17
Add click documentation

Cimbali:
This is a PR adding the docs for the python library [click](https://click.palletsprojects.com/en/8.1.x/). It only has 5 upvotes on the Trello board but I find myself using it offline quite regularly and annoyed at how little the doc is generally available.

The doc is in 2 main parts, “Documentation” (several pages / sections, which map well to types / entries) and “API” (single page, with sections containing functions/classes/etc.). I’ve made sure all these are searchable by making API sections types and API definitions entries. The remaining pages are listed as entries with as type the section under which they appear on the index page (currently only “Miscellaneous pages”).

License is open-sourced (BSD-3), but clause 3 specifies ”Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.” so I’ve added the following (similar to other docs) caveat to attribution to make sure we don’t claim endorsement/promotion:
> We are not supported nor endorsed by Pallets.


simon04:
The old versions fail to build, thus I'm removing them for now:

```
ERROR:                                                                                                                                                                                              
  https://click.palletsprojects.com/en/6.x/index.html
  Docs::ContainerFilter::ContainerNotFound: element '.body > section' could not be found in the document, url=https://click.palletsprojects.com/en/6.x/index.html
```

simon04 (approved):
Great, thank you!

PR #2256, opened 2024-06-04
Update Zig documentation (0.12.0)

katrinleinweber:

- [~] Ensured `self.links` contains up-to-date urls if `self.links` is defined

Not 100% sure about that last part, because:

```
$ bundle exec thor docs:generate zig@0.12.0
ERROR: could not find version "0.12.0" for doc "zig".
```

but: 1) This worked instead

```
$ bundle exec thor docs:generate zig --force
bundle exec rackup
Entries: (30 KB, +0 KB)
  Updated: 303
  Created: 0
  Deleted: 0
Types:
  Updated: 54
  Created: 0
  Deleted: 0
Files: (0.6 MB, +0.0 MB)
  Updated: 1
  Created: 0
  Deleted: 0
Done
```

and 2) This loads fine:

OTOH:

```
$ bundle exec thor docs:list | grep zig
zig
```

Is that because there is only 1 version of zig, as there is for many others?

```
bundle exec thor docs:list | grep -v '@' | wc -l
     128
```

I've followed 17d492b4856499831ff2ec32056f43ab6b55c7bb here.


simon04 (approved):
Thank you!

PR #2266, opened 2024-06-14
Update Elixir documentation (1.16 and 1.17)

breunigs:
The guides have moved from a separate page into the same documentation system. This required shuffling of the cleanup functions, which is subtly incompatible with the docs of the older versions, or would have required if/else in various places. For legibility, I thus opted to remove old versions.


simon04 (approved):
Thank you!

PR #2267, opened 2024-06-17
fixing yii scraper - the url scheme for the api docs changed (with everything else staying the same)

HayateLaTech:

### DIFF
--- a/lib/docs/scrapers/yii.rb
+++ b/lib/docs/scrapers/yii.rb
@@ -8,9 +8,8 @@ module Docs
     HTML
 
     version '2.0' do
-      self.release = '2.0.12'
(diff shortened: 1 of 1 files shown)


simon04:
I think this needs some more adaptions: The categories changed drastically. The guides are gone. The "x followers" on top of every pages needs to be removed.

devdocs io_yii_
localhost_9292_yii~2 0_

simon04 (changes requested):
I think this needs some more adaptions: The categories changed drastically. The guides are gone. The "x followers" on top of every pages needs to be removed.

devdocs io_yii_
localhost_9292_yii~2 0_

HayateLaTech:
I'll try to fix it! 😊💕

PR #2283, opened 2024-07-11
Update Vulkan version

morgabm:
Hello, so I recently discovered this tool with much excitement as Khronos Docs can sometimes be cumbersome to navigate. I was disappointed to learn the Vulkan version was only 1.0. I attempted to quickly update the Vulkan scraper to the new domain and version, but upon running `bundle exec thor docs:page vulkan` or any other command to pull/build/generate docs I get ssl cert error: 

`bundle exec thor docs:page vulkan`

I have verified the cert on the remote host. I also noticed an [old PR](https://github.com/freeCodeCamp/devdocs/pull/2117) suffered the same error due to changes in attribution methods. Since it has been 2 years I was wondering if something similar has changed causing these issues with the Vulkan scraper.

I attempted the same thor command above, with an unmodified version of the Vulkan scraper on the main branch, and it failed with the same error. Any suggestions?

- [X] Ensured the license is up-to-date

### DIFF
--- a/lib/docs/scrapers/vulkan.rb
+++ b/lib/docs/scrapers/vulkan.rb
@@ -2,11 +2,11 @@ module Docs
   class Vulkan < UrlScraper
     self.name = 'Vulkan'
     self.type = 'simple'
-    self.release = '1.0.59'
-    self.base_url = 'https://www.khronos.org/registry/vulkan/specs/1.0/'
-    self.root_path = 'apispec.html'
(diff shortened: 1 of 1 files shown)


simon04 (changes requested):
The scraper needs some adaptions. In particular, the extracted entries are completely different. All commands such as `vkCmdBindVertexBuffers` are missing from the entries now.

Screenshot 2024-07-17 at 06 13 38

PR #2291, opened 2024-07-26
fix rust 1.79.0 docs std crate indexing

### LINKED ISSUE #2290: Rust 1.79.0 docs missing content / unsearchable / parsing error
# Documentation style bug

## Summary
Rust docs contain many searchable elements like constants, functions, and traits.   For instance the standard library contains `std::f32::consts::PI`.

The current version of the docs has a parsing bug, where that page (and most of the rest of the pages in the `std` crate) has the short name `std::page`, with the title `Constant std::f32::consts::PICopy item path`

I suspect that the rust docs added a button to copy the item path to clipboard and that is breaking the existing  page scraping.

button tooltip

## Actual style

image

## Expected style

image

https://doc.rust-lang.org/stable/std/f32/consts/constant.PI.html

virtual256:
fixes #2290

### DIFF
--- a/assets/stylesheets/pages/_rust.scss
+++ b/assets/stylesheets/pages/_rust.scss
@@ -14,7 +14,7 @@
   em.stab, span.stab { @extend %label; }
   em.stab.unstable, span.stab.unstable { @extend %label-orange; }
   .out-of-band { float: right; }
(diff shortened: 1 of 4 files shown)


simon04 (approved):
Thank you!

PR #1441, opened 2020-12-23
Migrate CoffeeScript to JavaScript

simon04:
I'm migrating one after another CoffeeScript file to JavaScript using https://decaffeinate-project.org/repl/.

Fixes #1197.

PR #2307, opened 2024-08-19
Update README link and app description copy

djpowers:
- updates extension for Raycast extension (transferred ownership)
- removes an outdated reference to CoffeeScript

I've since taken over maintaining the DevDocs Raycast extension, so I updated the link to reflect this (the URL currently redirects, but this should be ensure longevity).

This also removes a references to using CoffeeScript, which was removed from the project in #1441.


simon04:
Ref: https://github.com/raycast/extensions/pull/14038

PR #2341, opened 2024-10-14
Fix Cordova indents with new clean html option

### LINKED ISSUE #1004: Cordova: Indentation Problem
In the Cordova 8 Doc of the cordova-file-plugin i saw intendation problems.

DevDoc-URL:
https://devdocs.io/cordova~8/reference/cordova-plugin-file/index#file-system-layouts

Original:

DevDocs:

thewheat:
Fixes #1004 

- Did this by adding a new option in `CleanHtmlFilter` and setting it in the Cordova scraper 
- The indents are removed as they are a non breaking space character that gets removed by CleanHtmlFilter. The option allows bypassing the whitespace removal if it detects a non breaking space
- I'm open to other ways of doing this more cleanly

Existing unindented with new indented
image

Source indented with new indented
image

### DIFF
--- a/lib/docs/filters/cordova/clean_html_core.rb
+++ b/lib/docs/filters/cordova/clean_html_core.rb
@@ -1,0 +1,11 @@
+module Docs
+  class Cordova
+    class CleanHtmlCoreFilter < Filter
+      def call
+        css('script', 'style', 'link').remove
+        xpath('descendant::comment()').remove
+        doc
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


simon04:
In order to not pollute the core filters with rarely used options, I've created `lib/docs/filters/cordova/clean_html_core.rb` which is used instead of `lib/docs/filters/core/clean_html.rb`

Updating cordova docs is blocked since the scraper needs to be adapted to changes on cordova.apache.org. I'll merge nevertheless.

PR #2365, opened 2024-11-19
fix: use aliases from config

ojeytonwilliams:
https://github.com/freeCodeCamp/devdocs/pull/2344 updated the aliases, but not everywhere. The help page still tried to access the old property and crashed.

Closes https://github.com/freeCodeCamp/devdocs/issues/2364


simon04:
Oliver, thank you for fixing this glitch!

ojeytonwilliams:
You're very welcome, Simon!

PR #2369, opened 2024-11-21
Update cmake up to v3.31 and add latest version

evakili:
This PR updates CMake docs up to version 3.31 and adds the latest docs as a separate entry. The CMake logo has also been updated to a newer one that Kitware uses.


simon04 (approved):
Thank you!

Issue #2330, opened 2024-09-20
Automate publishing Docker Images to DockerHub/GitHub Container Registry


ClariNerd617:
# Feature request

#2229  and #895 both ask for Docker images in Docker Hub.

## Summary

At this point in time, the official docker image for DevDocs is well over five years old, and thus pulling it makes me feel nervous. I would like to be able to self-host DevDocs, so it would make sense to add a GitHub Action to automatically build and push images from time to time. 

If necessary, I can fork this, figure this out, and open a PR, along with updated docs (such as for using `docker`/`podman` and/or `compose`.

ojeytonwilliams:
Now that https://github.com/freeCodeCamp/devdocs/pull/2386 is in, I think we can close this.

PR #2386, opened 2024-12-14
feat: add the GitHub action to create a Docker image and push to Docker Registry

chaitanyarahalkar:
This PR addresses #2330 

### Add automated Docker image builds

This PR adds a GitHub Actions workflow that automatically builds and publishes DevDocs Docker images to GitHub Container Registry (GHCR). The workflow addresses the concern of outdated official Docker images by providing fresh builds on a regular schedule.

### What this adds:
- New workflow .github/workflows/docker-build.yml that:
- Runs monthly on the 1st of each month
- Can be manually triggered via GitHub Actions UI
- Builds both regular and Alpine-based images
- Publishes to GHCR with the following tags:
  - ghcr.io/[org]/devdocs:latest
  - ghcr.io/[org]/devdocs:[YYYYMMDD]
  - ghcr.io/[org]/devdocs:alpine
  - ghcr.io/[org]/devdocs:alpine-[YYYYMMDD]
  
  The administrators for the repo would need to do the following - 
  
1. Enable GitHub Container Registry:
  - Go to repository "Settings" → "Actions" → "General"
  - Under "Workflow permissions", ensure "Read and write permissions" is selected
  - Save the changes
  - Enable Package Write Permissions:
2. Go to repository "Settings" → "Packages"
  - Ensure "Inherit access from source repository" is enabled
  - If not using inheritance, explicitly grant write access to GitHub Actions
3. Optional but recommended - Branch Protection
  - Go to repository "Settings" → "Branches"
  - Add branch protection rule for your main branch
  - Enable "Require status checks to pass before merging"
  - This ensures the Docker build workflow succeeds before merging
  
  
  I believe no other changes are needed here.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -15,10 +15,21 @@ jobs:
     permissions:
       contents: read
       packages: write
+    strategy:
+      matrix:
+        variant:
+          - name: regular
+            file: Dockerfile
+            suffix: ''
+          - name: alpine
+            file: Dockerfile-alpine
+            suffix: '-alpine'
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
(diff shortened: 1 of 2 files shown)


simon04:
Cool, thank you for woking on this feature. Some remarks:

- Please update the README for usage instructions ("Alternatively, DevDocs may be started as a Docker container"; maybe move the Docker instructions up since they are much simpler?)
- Steps for standard and alpine are very much alike. Have you considered using [`matrix`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow)? Does it make sense here?
- The audit tool [zizmor](https://github.com/woodruffw/zizmor?tab=readme-ov-file) outputs one warning (see below)

@ojeytonwilliams, I would love to hear your thoughts on this PR. 😄 

```
> uvx zizmor .github/workflows/docker-build.yml 
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> /Users/simon/src/devdocs/.github/workflows/docker-build.yml:16:9
   |
16 |         - name: Checkout repository
   |  _________-
17 | |         uses: actions/checkout@v4
   | |_________________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low
```

ojeytonwilliams:
It looks like this is a good approach, but I noticed a couple of strange things.

 "Settings" → "Packages" doesn't seem to exist. Perhaps it won't until we publish the first one, I'm not sure.
 
Branch protection rules don't seem relevant when we're running this via workflow dispatch or cron.

Other than that (and what Simon pointed out), I think this all makes sense.

chaitanyarahalkar:
@ojeytonwilliams I believe that setting change is not needed to be done. It should work out of the box..

@simon04 added the matrix option for the build image.. I believe the warning should go as well since added the option to not persist credentials. Updated the README as well. 

Thanks for your comments!

simon04 (approved):
Great! Let's give it a try. 😄

ojeytonwilliams:
And there it is: https://github.com/freeCodeCamp/devdocs/pkgs/container/devdocs

Cheers!

Issue #2351, opened 2024-11-02
Feature Request: Add Support for Three.js or Documentation on Integration


tomdevisser29:
Thank you so much for creating this, it's really valuable. I would love to see Three.js added to the docs, though! Is that possible? If I can contribute by adding it myself, that's okay too, please let me know how.

First, thank you for developing this project — it's incredibly useful! I wanted to request the addition of Three.js to the documentation. Is there a possibility of incorporating this library or providing guidance on how it can be integrated?

If contributing by adding this myself is an option, I’m more than willing to help. Please let me know the best way to proceed or any relevant contribution guidelines.

Thank you!

simon04:
Fixed via https://github.com/freeCodeCamp/devdocs/pull/2392

PR #2392, opened 2024-12-24
feat(documentation): Add documentation support for three.js

chaitanyarahalkar:
- [x] Added these files to the <code>public/icons/*threejs*/</code> directory:

Addressed issue #2351

This PR adds documentation support for `three.js`. The parsers for this were slightly complicated since `three.js` uses a different mechanism to render its documentation. The `FileScraper` based approach is used where the documentation is first downloaded and then the parser is applied on the locally downloaded files.


simon04 (approved):
Great, thank you!

spamguy:
Closes #2351.

PR #2414, opened 2025-01-21
fix: prevent compression with Accept-Encoding

ojeytonwilliams:
It seems that (apparently randomly) the index.json files are getting compressed and causing URI.open to fail. Possibly, it's a little unclear.

Nonetheless, sending the "Accept-Encoding" of "identity" seems to do the trick.


ojeytonwilliams:
@simon04 I don't really get why this was necessary. I haven't changed anything over the last few days.

Let me know if it starts failing again, but hopefully this is enough.

PR #2439, opened 2025-02-21
fix: open external links in default browser from macOS PWA

cpmsmith:
By using `window.open(url, "_blank", "noopener")` directly instead of a very old workaround, Safari can detect before opening the new window that it's going to an external URL, and send the link to the default browser. I tested this locally, and although MDN claims that, on iOS, ["this method will not function if the `target` parameter is unspecified or set to `_blank`"](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#browser_compatibility), in my testing it seemed to work fine.

Fixes freeCodeCamp/devdocs#2438

### DIFF
--- a/assets/javascripts/lib/util.js
+++ b/assets/javascripts/lib/util.js
@@ -457,13 +457,13 @@ $.noop = function () {};
 
 $.popup = function (value) {
   try {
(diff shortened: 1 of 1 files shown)


simon04 (approved):
Great, thank you!

PR #2440, opened 2025-02-21
Update Tensorflow and Tensorflow C++ (2.16)

ClasherKasten:

- [X] Ensured the license is up-to-date

### DIFF
--- a/lib/docs/filters/tensorflow/clean_html.rb
+++ b/lib/docs/filters/tensorflow/clean_html.rb
@@ -46,6 +46,9 @@ module Docs
           node.replace("<p>#{node.to_html}</p>")
         end
 
+        css('span[slot="popout-heading"]').remove
+        css('span[slot="popout-contents"]').remove
+
         doc
       end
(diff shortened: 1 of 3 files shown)


simon04 (approved):
Thank you!

ClasherKasten:
@simon04 I don't think this should've been bumped to `2.18` as there is no such version on the [tensorflow.org](https://tensorflow.org/versions) website. If the site used to scrape the version added by this PR is opened, it displays version `2.16.1`.

It could be labeled as `2.17` though i think because the links of `2.17` and `2.16` point to the same page (the ones specified in the scraper) and the changelog also suggests that this is the documentation for the `2.17.x` release.

simon04:
Thanks for pointing out the version. I've obtained 2.18 using `bundle exec thor updates:check tensorflow` which checks for the latest release on https://github.com/tensorflow/tensorflow/releases. I'll fix `get_latest_version` and upload version 2.16.1

PR #2441, opened 2025-02-21
Fix multiple version checks of scrapers.

ClasherKasten:
This PR fixes the following version checks:
- Docker
- Meteor
- Nix
- PHPUnit
- Redis
- Sass
- Trio


simon04 (approved):
Great, thank you!

PR #2469, opened 2025-03-31
Remove the "::" separator from the SEPARATORS_REGEXP in searcher.js 

### LINKED ISSUE #2463: Searching C++ doc by URL is broken
# Bug report

## OS information
Tested using:
- Windows Chrome 134.0.6998.89
- Android Chrome 134.0.6698.96
- Windows Firefox 136.0.1
- Android Firefox 136.0.2

## Steps to reproduce
Enter URL in browser:
https://devdocs.io/#q=c++%20std::min

According to http://devdocs.io/help#search, that should do a search in the C++ doc for `std::min`.  Instead, it searches the C++ doc for `  std::min`, with two leading spaces, and shows poor quality search results.  Manually removing those leading spaces in the search box causes good search results to come up again.

I've used this feature for keyword searches from my browsers' URL bar for a long time, except as `https://devdocs.io/#q=cpp%20%s` (`cpp` rather than `c++`), but starting very recently, that no longer selects the doc I intended.

## Possible fix
Trim leading spaces on C++ search terms from the URL!

Alternatively, it seems like it would be slightly cleaner if single-doc searches could be done with URLs like:
https://devdocs.io/cpp/#q=std::min

chaitanyarahalkar:
Should fix #2463 

Tried searching for `std::min` and was able to get good C++ documentation results.

### DIFF
--- a/assets/javascripts/app/searcher.js
+++ b/assets/javascripts/app/searcher.js
@@ -154,7 +154,7 @@ app.Searcher = class Searcher extends Events {
   };
 
   static SEPARATORS_REGEXP =
-    /#|::|:-|->|\$(?=\w)|\-(?=\w)|\:(?=\w)|\ [\/\-&]\ |:\ |\ /g;
+    /#|:-|->|\$(?=\w)|\-(?=\w)|\:(?=\w)|\ [\/\-&]\ |:\ |\ /g;
   static EOS_SEPARATORS_REGEXP = /(\w)[\-:]$/;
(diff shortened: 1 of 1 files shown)


chaitanyarahalkar:
@simon04 let me know if this change looks good. Ty!

simon04:
Hi, thank you for this PR. I have troubles understanding the practical benefit when not considering `::` as a separator. The given example `std::min` works just fine at the moment? Are there other search inputs that benefit from removing `::` as a separator?

PR #2490, opened 2025-05-02
switch neovim plugin link to a newer maintained plugin

emmanueltouzery:
unfortunately the neovim devdocs integration plugin that's linked from the readme was archived on github. This PR switches the link to a new similar plugin (that I wrote) that's maintained.


simon04 (approved):
Thank you!

PR #2494, opened 2025-05-13
Crystal 1.16.3

spuun:


simon04:
Hi, thanks for creating this PR, but this not a change we will be making at this time. We stick to stable releases and refrain from patch version updates (unless they provide significant value).

According to https://github.com/crystal-lang/crystal/releases the releases 1.16.2 and 1.16.3 contain only a few bugfixes.

PR #2513, opened 2025-05-31
Update pytorch scraper, include various 2.x versions

blahgeek:
1. Various 2.x versions are included separately. Pytorch versions are not backward compatible, it has different compatibilities between CUDA etc, so people may use specific versions for a extended period of time.

2. Removed the type replacement table for `get_type`. Instead, get the type from breadcrumbs directly. IMO this produces better results that matches the index in the original website (the left side menu in docs.python.org). Also, the `TYPE_REPLACEMENT` table was opiniated and hard to maintain across versions.

3. Always include default entry (removed `include_default_entry?` function). I don't see the downside of this. Previously some pages are missing because of this (e.g. torchrun https://docs.pytorch.org/docs/1.13/elastic/run.html)

### DIFF
--- a/lib/app.rb
+++ b/lib/app.rb
@@ -376,6 +376,8 @@ class App < Sinatra::Application
     'angular~1.3' => 'angularjs~1.3',
     'angular~1.2' => 'angularjs~1.2',
     'codeigniter~3.0' => 'codeigniter~3',
+    'pytorch~1' => 'pytorch~1.13',
+    'pytorch~2' => 'pytorch',
     'webpack~2' => 'webpack'
   }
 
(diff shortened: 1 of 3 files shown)


blahgeek:
cc @ArkciaTheDragon for modifying some code from https://github.com/freeCodeCamp/devdocs/pull/2137

ArkciaTheDragon:
Tested on PyTorch 2.7 — looks cool!

Also +1 to the suggestion of using separate PyTorch versions. For context: PyTorch 2.7 generates only 16.2MB of offline data, which is actually a bit smaller than Python 3.12’s ~18MB. I hope hosting the versions separately won’t be an issue in terms of space.

simon04 (approved):
Thank you!

PR #2516, opened 2025-06-02
Add get_latest_version for opengl

Suraj-Yadav:

Also updated base_url's, previous ones didn't exist.

### DIFF
--- a/lib/docs/scrapers/opengl.rb
+++ b/lib/docs/scrapers/opengl.rb
@@ -27,5 +27,9 @@ module Docs
       self.release = '2.1'
       self.base_url = "https://registry.khronos.org/OpenGL-Refpages/gl#{self.version}/"
     end
+
+    def get_latest_version(opts)
+      get_latest_github_commit_date('KhronosGroup', 'OpenGL-Refpages', opts)
+    end
   end
 end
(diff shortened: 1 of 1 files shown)


simon04:
Currently, the output for the version check is

```
> bundle exec thor updates:check opengl
+--------------------------------------------------+
|    Outdated major version documentations (1)     |
+---------------+-----------------+----------------+
| Documentation | Scraper version | Latest version |
+---------------+-----------------+----------------+
| OpenGL        | 4               | 2024-11-20     |
+---------------+-----------------+----------------+
```

How is the commit date intended to relate to the scraper version?

Suraj-Yadav:
I need some clarity on this point.
OpenGL only has api reference for the latest version (currently 4.5), under gl4. This is `opengl~4` in devdocs. This gets [updated](https://github.com/KhronosGroup/OpenGL-Refpages/commits/main/gl4) and I want this to be captured in the Documentation versions reports.

OpenGL also have api reference for OpenGL 2.1, under gl2.1. This is `opengl~2.1` in devdocs. These are retained because they are the only source of reference material for GLX, GLU, and the OpenGL Compatibility Profile.

Ref: https://registry.khronos.org/OpenGL-Refpages/

What would be the best way to have this?

PR #2536, opened 2025-07-10
i added some words

Ranjankrs:
this is way to understand fork


simon04:
This is not how GitHub works! Stop spamming open-source projects, please!

Issue #2529, opened 2025-06-23
RxJS does not include top level "index" functions


forivall:
# Documentation style bug

## Summary

The RxJS docs are missing the top level "index" functions.

## Actual style

Only ajax, fetch, operators, testing and webSocket sections are included.

<img width="322" alt="Image" src="https://github.com/user-attachments/assets/b183648a-444b-42b8-85ca-d879ecb8e71a" />

## Expected style

<details>
<summary>"index" section is included in official docs, in addition to the above sections.</summary>

![Image](https://github.com/user-attachments/assets/68b78a6a-3876-49df-a44e-1e1729ce3d1f)

</summary>

aditya-7562:
Thank you for reporting this issue, @forivall .

After investigating the RxJS doc structure, I believe the root cause lies in the current DevDocs scraper setup:

1. **Scraper Skipping `/api/index`**:  
   The `rxjs.rb` scraper currently has the pattern `/api/index/` listed in `options[:skip_patterns]`, which prevents any of the top-level functions (like `of`, `from`, `interval`, etc.) from being fetched or processed. I plan to remove this from the skip list and explicitly include `generated/docs/api/index.json` (or its children) in the `initial_urls` to ensure these entries are scraped.

2. **Sidebar Grouping in `entries.rb`**:  
   Even after scraping, functions from `/api/index/function/...` wouldn't be labelled or grouped properly in the sidebar. In the RxJS-specific `entries.rb` filter, I'll enhance:
   - `get_name` to append `()` to function names even if `.api-type-label.function` is missing (which is the case for many index functions).
   - `get_type` to classify these under a new `"Top-level Functions"` section for better sidebar clarity.

Once the scraper and filter are updated, functions such as `of()`, `from()`, etc., will appear and be searchable in the DevDocs RxJS documentation.

Let me know if you're open to a PR addressing this!

simon04:
Closing, see discussion in #2540.

forivall:
I'd need to do some analysis to verify, but i think everything in `operators` is also in `index` - so, the solution could either drop a separate `operators` section and include only the `index`, or dedupe and remove the items which are in `operators` from the `index` section. Either way, this would be an improvement, because as it stands, things like `asyncScheduler` and `of` and `from` are not available on DevDocs.

PR #2540, opened 2025-07-17
fix(rxjs): Include index functions and improve scraper coverage (#2529)

aditya-7562:
Fixes [#2529](https://github.com/freeCodeCamp/devdocs/issues/2529)

This PR enhances the RxJS documentation by introducing a dedicated Top-level Functions category in the DevDocs sidebar. This improves discoverability for core RxJS functions such as of(), from(), and merge(), which were previously uncategorized or grouped under "Miscellaneous".

✅ Changes Made:

- Updated `lib::docs::filters::rxjs::entriesget_type `to return `Top-level Functions` for root-level `api/` entries that don't belong to any operator class (e.g., `ajax`, `animationFrameScheduler`, etc.).

- Sidebar now reflects the new grouping for standalone functional APIs.

- Ensured no impact to existing guide sections or categorized operators.

🧪 Testing

-  Ran the scraper locally and verified a successful build

-  Confirmed sidebar now contains a Top-level Functions section

-  Verified visual consistency and link correctness across updated entries

### DIFF
--- a/lib/docs/filters/rxjs/entries.rb
+++ b/lib/docs/filters/rxjs/entries.rb
@@ -2,18 +2,28 @@ module Docs
   class Rxjs
     class EntriesFilter < Docs::EntriesFilter
       def get_name
-        title = at_css('h1')
-        name = title.nil? ? subpath.rpartition('/').last.titleize : title.content
-        name.prepend "#{$1}. " if subpath =~ /\-pt(\d+)/
-        name += '()' unless at_css('.api-type-label.function').nil?
+        # Get function name from path if it's inside api/index/function
+        if subpath.start_with?('api/index/function/')
+          name = subpath.split('/').last.sub(/\.html$/, '').titleize
+          name += '()'
+        else
+          title = at_css('h1')
+          name = title.nil? ? subpath.rpartition('/').last.titleize : title.content
+          name.prepend "#{$1}. " if subpath =~ /\-pt(\d+)/
+          name += '()' unless at_css('.api-type-label.function').nil?
+        end
         name
       end
 
       def get_type
         if slug.start_with?('guide')
           'Guide'
+        elsif slug == 'api/index'
+          nil # Hide index page to avoid listing the full page in sidebar
+        elsif slug.start_with?('api/index/function/')
+          'Top-level Functions'
         elsif slug.start_with?('api/')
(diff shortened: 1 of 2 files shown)


simon04:
timeInterval appears twice now as http://localhost:9292/rxjs/api/operators/timeinterval and http://localhost:9292/rxjs/api/index/function/timeinterval. Is this intended?

aditya-7562:
> timeInterval appears twice now as http://localhost:9292/rxjs/api/operators/timeinterval and http://localhost:9292/rxjs/api/index/function/timeinterval. Is this intended?

Thanks for catching that!

Yeah, the duplication is actually intentional — it matches how the official RxJS docs list timeInterval in both the standalone functions and the operators section. Since it’s validly part of both, the scraper now reflects that to stay consistent with the source.

- https://rxjs.dev/api/operators/timeInterval
- https://rxjs.dev/api/index/function/timeInterval

That said, if you think we should clean it up or consolidate one of the paths in DevDocs, I’m happy to make the change!

aditya-7562:
Hey @simon04  Just checking in on this PR.

As you pointed out, `timeInterval` shows up in both the operators and top-level functions sections — that’s actually intentional to match the official RxJS docs, which list it in both places. The idea was to stay consistent with that structure, but I’m totally open to changing it if you think it makes more sense to consolidate.

Would love to hear your thoughts when you get a chance — happy to tweak things as needed!

simon04:
Many users of devdocs simply use the search to find symbols when they already know the library. I dislike that every function is shown twice in the results (with different typesetting) when the actual content is indistinguishable.

For comparison, searching for timeInterval on rxjs.dev only results in https://rxjs.dev/api/index/interface/TimeInterval and https://rxjs.dev/api/index/function/timeInterval

image

aditya-7562:
Thanks for the thoughtful feedback, @simon04  — that’s a really good point about the search results feeling cluttered with duplicates.

I included `timeInterval` in both sections to stay close to how the official docs list is, but I totally get that DevDocs has a different focus, especially when it comes to usability and search clarity.

Would you prefer we just keep it under `operators/` or under `index/function/`? I’m happy to clean it up either way — just want to make sure it fits well with how DevDocs is organized.

Appreciate the guidance!

Issue #2568, opened 2025-09-05
Rust Reference page title issues


cpmsmith:
# Documentation style bug

## Summary
A few different things around page titles for The Rust Reference:

1. All page titles in the nav begin with `.. `
2. Page titles in the nav don't match the title when viewing them, e.g. [Grouped expressions](https://devdocs.io/rust/reference/expressions/grouped-expr) shows `.. The Rust Reference`, and [Panic](https://devdocs.io/rust/reference/panic) shows `.. The panic_handler attribute`
3. Anchor IDs, e.g. `[expr.paren]` show up as headings before the actual page heading

May be related to #2510 or its fix

## Actual style
Many pages, including: https://devdocs.io/rust/reference/expressions/grouped-expr

<img width="1357" height="972" alt="Image" src="https://github.com/user-attachments/assets/17277f73-03a5-4158-86b1-01c13047bc25" />

## Expected style

https://doc.rust-lang.org/reference/expressions/grouped-expr.html#r-expr.paren

<img width="1270" height="879" alt="Image" src="https://github.com/user-attachments/assets/0c3f3aa9-13d2-4b60-8e23-d92bd49f4fc9" />

cpmsmith:
Oh also, a couple of pages in the Rust Guide ("the book") have the same `.. ` problem, which I think messes up the sorting within the Guide (note 21.03 above 00.00)

<img width="325" height="928" alt="Image" src="https://github.com/user-attachments/assets/9ceea6bc-dfe6-4d9d-9d74-2532b065a30a" />

spamguy:
~I can work on this.~

cpmsmith:
@spamguy 
> I can work on this.

I actually got around to it in #2569. However there is still some unusual stuff going on with the `std::` stuff as well, where almost everything is getting put in the `std` type, even when it should be in some more specific module, e.g. `std::process::abort` and `std::intrinsics::abort` are both labelled `std::abort`, meanwhile `std::intrinsics` is empty:

<img width="122" height="137" alt="Image" src="https://github.com/user-attachments/assets/f9ac1816-705c-4e6b-b226-9299b5ae1387" />

<img width="320" height="121" alt="Image" src="https://github.com/user-attachments/assets/3f9e7330-b953-4c35-a248-3dbb8b9cdd77" />

spamguy:
Nice. I'll leave you to it then.

PR #2569, opened 2025-09-08
Improve Rust scraper

cpmsmith:
Several improvements to the Rust scraper here. In screenshots, before is on the left, after is on the right.

## Improve broken page names in Rust
As described in #2568:
- all pages in the Rust Reference had `.. ` on the beginning of their title
- many had their title as `.. The Rust Reference`
- others had their title as the first subheading on the page, but not the main title
- and some in The Book had similar problems

Now, no empty dots, things in the intro of the book all get `00.00.` so they're sorted to the top, and pages in the Book and the Reference all get the correct title.

<details>
<summary>Screenshots</summary>
image

image
</details>

## Update to new compiler error codes location
It got moved, and is no longer all on one page.

<details>
<summary>Screenshot</summary>
image
</details>

## Fix and style syntax definitions/diagrams in Rust Reference
The Rust Reference has syntax diagrams and definitions for a bunch of things, which were missing styles and elements before, and had a non-functional toggle button. I put them in a working `<details>` element and adapted SVG styles from the original.

<details>
<summary>Screenshot</summary>
image
</details>

> [!NOTE]
> This meant changing the global `CleantTextFilter` to include some SVG elements if they're empty. This affects all the docs, but I think in a mostly positive way. For instance, the Web APIs docs have some broken SVGs in them for this same reason:
> <details>
> <summary>Screenshot</summary>
> image
> </details>
>
> …however, I haven't verified that this will improve this exact case, because the MDN scraper appears to be broken at the moment. (It's broken on main as well)
>
> I also made some modifications to the global link handling, because:
> 1. `a` tags in an SVG don't behave quite like normal ones, so they would break history navigation
> 2. Unlike native browser navigation, navigating to a fragment inside a closed `<details>` element didn't open the `<details>`, and I added a lot of those, so I implemented that.

## Remove visible anchor links from Rust Reference
These are hidden in the original docs, and they're all over the place:

<details>
<summary>Screenshot</summary>
image
</details>

### DIFF
--- a/lib/docs/scrapers/rust.rb
+++ b/lib/docs/scrapers/rust.rb
@@ -3,7 +3,7 @@
 module Docs
   class Rust < UrlScraper
     self.type = 'rust'
-    self.release = '1.88.0'
+    self.release = '1.90.0'
     self.base_url = 'https://doc.rust-lang.org/'
     self.root_path = 'book/index.html'
     self.initial_paths = %w(


assets/javascripts/lib/util.js
```diff
@@ -353,6 +353,16 @@ $.lockScroll = function (el, fn) {
...
+// If `el` is inside any `<details>` elements, expand them.
+$.openDetailsAncestors = function (el) {
+  while (el) {
+    if (el.tagName === "DETAILS") {
+      el.open = true;
+    }
+    el = el.parentElement;
+  }
+}
+
```
simon04, line 365: Please do not change this globally since `<details>` is used in many docs. 

If `<details>` should be opened for Rust, add an `open` attribute via `lib/docs/filters/rust/clean_html.rb`

PR #2604, opened 2025-10-17
`#contributors` chat room in Discord

denilsonsa:
According to ILM, "less used channels were merged". Thus, there is no longer a `devdocs` channel. Instead, we should join `#contributors`.


ojeytonwilliams (approved):
Thanks @denilsonsa, LGTM 👍

Issue #2635, opened 2025-12-30
Docker build failure due to bundler flags


hugginsio:
# Bug report

The Docker image fails to build using the `docker build -t devdocs .` command, perhaps due to Bundler flags:

```
[+] Building 1.5s (9/11)                                                                                                                                    docker:orbstack
 => [internal] load build definition from Dockerfile                                                                                                                   0.0s
 => => transferring dockerfile: 529B                                                                                                                                   0.0s
 => [internal] load metadata for docker.io/library/ruby:3.4.7                                                                                                          1.1s
 => [internal] load .dockerignore                                                                                                                                      0.0s
 => => transferring context: 104B                                                                                                                                      0.0s
 => [1/7] FROM docker.io/library/ruby:3.4.7@sha256:779ca68b84fe3804e27f5b7ea5ab001005c708b7236a798c15b860d71cba9d21                                                    0.0s
 => [internal] load build context                                                                                                                                      0.1s
 => => transferring context: 118.44kB                                                                                                                                  0.1s
 => CACHED [2/7] WORKDIR /devdocs                                                                                                                                      0.0s
 => CACHED [3/7] RUN apt-get update &&     apt-get -y install git nodejs libcurl4 &&     gem install bundler &&     rm -rf /var/lib/apt/lists/*                        0.0s
 => CACHED [4/7] COPY Gemfile Gemfile.lock Rakefile /devdocs/                                                                                                          0.0s
 => ERROR [5/7] RUN bundle install --system &&     rm -rf ~/.gem /root/.bundle/cache /usr/local/bundle/cache                                                           0.3s
------
 > [5/7] RUN bundle install --system &&     rm -rf ~/.gem /root/.bundle/cache /usr/local/bundle/cache:
0.224 The `--system` flag has been removed because it relied on being remembered
0.224 across bundler invocations, which bundler no longer does. Instead please use
0.224 `bundle config set path.system true`, and stop using this flag
------

 1 warning found (use docker --debug to expand):
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 24)
Dockerfile:14
--------------------
  13 |
  14 | >>> RUN bundle install --system && \
  15 | >>>     rm -rf ~/.gem /root/.bundle/cache /usr/local/bundle/cache
  16 |
--------------------
[...]

PR #2636, opened 2025-12-30
fix(docker): address `system` and `without` flag deprecations in Bundler

hugginsio:
I have updated the Dockerfiles to address the deprecation issue that results in the build failures as reported in #2635.

Both images build, run, and otherwise function as I would expect.

Please let me know if any additional testing is required. Thanks!


hugginsio:
Thanks, Oliver!

PR #2665, opened 2026-03-18
Add ruby/rack docs

thedanbob:
This PR adds docs for the Rack library, which is arguably the most important Ruby library when it comes to web development. Anyone who regularly uses Rails or one of the other Ruby web frameworks will likely need to consult the Rack docs at some point.

------


simon04:
Hi ,thanks for contributing. I'm not familiar with this library, but I would expect to find https://rack.github.io/rack/3.2/Rack/QueryParser.html when searching for "queryparser", but I only find two other results:

image

thedanbob:
Whoops, I missed one of the scraper reference steps. Both this one and minitest don't generate properly on rdoc 7, in particular there's no table_of_contents.html anymore and the HTML structure is different. I guess the rdoc scraper will have to be overhauled if/when it's no longer possible to use rdoc <7.

simon04 (approved):
Awesome, thank you!

Issue #2590, opened 2025-10-04
Update Terraform docs (0.11.7 → 1.13.3)


Luis-Lpz:
# Update Terraform docs (0.11.7 → 1.13.3)

## Summary
Terraform documentation in DevDocs is outdated.  
Currently it shows version **0.11.7**, while the latest stable release is **1.13.3**.  

## Actual style
DevDocs Terraform docset:  
https://devdocs.io/terraform/

It only includes Terraform 0.11.7 docs.

## Expected style
Latest Terraform docs (1.13.3) from HashiCorp:  
https://developer.hashicorp.com/terraform/docs

Please update the Terraform docset to the latest version.

spamguy:
I gave a go at this and ran into rate limiting almost instantly through `UrlScraper`:

```
Queue:   developer.hashicorp.com/terraform/language
ERROR:
  https://developer.hashicorp.com/terraform/language
  RuntimeError: Error status code (429): No error
    https://developer.hashicorp.com/terraform/language
    cache-control":"private, no-store, max-age=0","content-type":"text/html; charset=utf-8","server":"Vercel","x-vercel-challenge-token":"2.1760290904.60.Mz...","x-vercel-id":"sfo1::17...","x-vercel-mitigated":"challenge","date":"Sun, 12 Oct 2025 17:41:44 GMT
```

simon04:
> and ran into rate limiting almost instantly

You might want to use `options[:rate_limit]`

spamguy:
Good to know. Unfortunately that didn't help, even with a value of 1. Either it knows how to sniff our URL scraper, or the process, which I rebuilt from the ground up, is altogether misconfigured.

PR #2675, opened 2026-04-03
New docs: Opentofu

5long:
Add docs for OpenTofu: a fork of Terraform since upstream changed its license to BUSL.

The current Terraform docs is out-of-date (#2590). OpenTofu can serve as a decently compatible replacement (see <https://cani.tf>). Some might argue that it's preferred than the upstream due to community-driven development.

### Known Issue

At least one page `http://localhost:9292/opentofu~1.11/cli/config/environment-variables/index` is truncated at around `TF_LOG_PATH` for no apparent reason. [Upstream page](https://opentofu.org/docs/cli/config/environment-variables/) is fairly big -- the raw HTML is 98k. However, there are bigger pages (e.g. http://localhost:9292/opentofu~1.11/language/settings/backends/s3/index , 104k HTML) that doesn't get truncated. I'm still trying to figure out why. But I guess imperfect docs is better than no docs at all.

### Misc

- Only officially support versions 1.9 ~ 1.11 are included & tested. Unsupported versions 1.6 ~ 1.8 can be found on opentofu.org which can be added if needed.
- To keep things simple, the code treats `Opentofu` as a single word without having to capitalize `t`.


simon04 (approved):
Great, thank you!

PR #2681, opened 2026-04-25
Update Deno documentation scraper

srpatcha:


simon04:
The scraper needs fine tuning:

image

image

srpatcha:
Thanks for the review and the screenshots @simon04! You're right that the output needs refinement.

From the screenshots I can see a few issues to address:
1. Residual navigation/UI chrome that isn't being stripped by the current `clean_html` selectors
2. Code block language detection could be more accurate

I'll iterate on the `clean_html.rb` and `entries.rb` filters to clean up the rendering. Could you confirm which Deno docs section/URL the screenshots were captured from so I can target the exact page structure when tuning the selectors? That will help me verify the fixes against the same pages you're seeing.

simon04 (changes requested):
.

srpatcha:
@simon04 I pushed d17aafb6 to address the requested scraper fine-tuning without waiting for the missing source URL. I matched the screenshots to the current pages and validated the previous vs updated filter on both:

- `https://docs.deno.com/api/deno/network/`: breadcrumbs `1 -> 0`, symbol-kind badges `64 -> 0`, code-copy buttons `23 -> 0`, feedback sections `1 -> 0`; all 23 TypeScript blocks still emit `pre[data-language="ts"]`.
- `https://docs.deno.com/runtime/reference/std/fmt/`: copy-page controls `1 -> 0`, code-copy buttons `2 -> 0`, mobile “On this page” blocks `1 -> 0`, heading anchors `1 -> 0`, feedback sections `1 -> 0`; JavaScript and shell blocks remain `js` and `sh`.

The filter now scopes output to `main#content article`, removes the confirmed source-site chrome, and transfers each live `language-*` class to the enclosing `pre[data-language]` instead of defaulting every block to TypeScript. I also added reduced API/runtime fixtures and filter/entry regressions.

Validation:
- Focused Deno tests: 17 runs, 73 assertions, 0 failures/errors.
- Remaining Ruby suite excluding two independently reproduced Windows-only baseline files: 665 runs, 927 assertions, 0 failures/errors.
- Full suite reached 761 runs; only the existing Windows case-sensitivity URL assertion and FileStore `test/tmp` path errors failed when run independently.
- Ruby syntax checks, editor diagnostics, change validation, and `git diff --check` passed.
- The live Thor scraper reached both exact URLs; local Windows curl then stopped at its CA-path configuration, so the before/after output comparison used the already-downloaded current HTML through the real DevDocs parser/filter stack.

Re-review is requested because the two issues identified from the screenshots are now covered directly.

simon04 (approved):
Thank you!

Issue #2632, opened 2025-12-05
Add Odin


Crenshinibon:
# Feature request

It would be nice to also have the odin documentation here. And while you are on it adding raylib and the other odin/vendor tooling would be great as well.

https://odin-lang.org/

simon04:
Fixed via https://github.com/freeCodeCamp/devdocs/pull/2687.

PR #2687, opened 2026-05-26
Add Odin documentation

jask06:

NOTE: No 16@2x.png image was available from the Odin repo, so I used `https://convertico.com/favicon-resizer/` to upscale the available 16x16 image.

This should also satisfy #2632 

Please let me know your thoughts!


jask06:
Please let me know if there are any issues with the current PR!

lib/docs/scrapers/odin.rb
```diff
@@ -0,0 +1,38 @@
...
+          docs/nightly
+          docs/odin-book
+          docs/spec
+          docs/packages
+        )
+
+        options[:attribution] = <<-HTML
+          &copy; 2016-#{Date.today.year} Ginger Bill<br>
+          Licensed under the 3-clause BSD License.
+        HTML
```
simon04: The license has been changed, see https://github.com/odin-lang/Odin/commit/842cfee0f3eacbef5fa5f0dbd392a500ebad93ae.

lib/docs/scrapers/odin_packages.rb
```diff
@@ -0,0 +1,28 @@
+module Docs
+    class OdinPackages < UrlScraper
+        self.name = 'Odin Packages'
+        self.slug = 'odin_packages'
+        self.type = 'odin_packages'
```
simon04: I would prefer to have this separate scraper to be merged with the main `odin`. Using one without the other does not seem overly helpful to me. Also, this follows practice we've been following for other scrapers (such as `crystal` or `rust`).

jask06: Thank you! This makes a lot of sense. When I was looking, I didn't recognize there were parsers with multiple URLs that could be executed in the same scrape (my own skill issue). Thank you for pointing that out! 

I'll look at those and get that updated!

lib/docs/scrapers/odin.rb
jask06: Got it! I'll get this updated!

PR #2688, opened 2026-05-26
fix: retry failed downloads

ojeytonwilliams:
This makes the build retry failed downloads a couple of times. 

I checked locally and it handles the failures without any obvious issues.

Ref: https://github.com/freeCodeCamp/devdocs/issues/2686

### DIFF
--- a/lib/tasks/docs.thor
+++ b/lib/tasks/docs.thor
@@ -235,6 +235,7 @@ class DocsCLI < Thor
     puts 'Docs -- BEGIN'
 
     require 'open-uri'
+    require 'net/http'
     require 'thread'
 
     docs = Docs.all_versions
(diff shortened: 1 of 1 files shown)


simon04 (approved):
Amazing, thank you for implementing this workaround.

ojeytonwilliams:
No worries, it looks like it was enough.

It's a little hard to figure out what's wrong because it seems the bucket sometimes simply doesn't respond. I'll see if I can find out why.

ojeytonwilliams:
Rather curiously, it only needed to retry once in the latest deploy.

simon04:
Deployments used to take only approx 2min in 2025-12; since then the times have steadily increased: https://github.com/freeCodeCamp/devdocs/actions/workflows/build.yml

ojeytonwilliams:
Right. We could be seeing failures more often simply because the number of requests keeps growing. Maybe. I'm not 100% sure that's the whole story, though.

PR #2695, opened 2026-06-04
Fix plus decoding in URL search hashes

### LINKED ISSUE #2463: Searching C++ doc by URL is broken

puneetdixit200:
## Summary

- add a hash-fragment decoder that preserves literal `+` characters
- use it for URL search hash parsing in both `Search` and `SearchScope`
- cover scoped C++ URL searches and encoded literal plus queries with a focused Node test

Fixes #2463.

## Tests

- `node --test test/assets/search_hash_test.js`
- `node --check assets/javascripts/lib/util.js && node --check assets/javascripts/views/search/search.js && node --check assets/javascripts/views/search/search_scope.js && node --check test/assets/search_hash_test.js`
- `git diff --check`

Not run locally: `bundle exec rake` because this checkout requires Ruby 4.0.5 and Bundler 4.0.12, while this machine only has system Ruby 2.6.10 and no matching Bundler.

### DIFF
--- a/assets/javascripts/lib/util.js
+++ b/assets/javascripts/lib/util.js
@@ -450,6 +450,9 @@ $.escapeRegexp = (string) => string.replace(ESCAPE_REGEXP, "\\$1");
 
 $.urlDecode = (string) => decodeURIComponent(string.replace(/\+/g, "%20"));
 
+// Hash fragments are not form-encoded, so literal plus signs must stay plus signs.
+$.urlDecodeFragment = (string) => decodeURIComponent(string);
+
 $.classify = function (string) {
   string = string.split("_");
   for (let i = 0; i < string.length; i++) {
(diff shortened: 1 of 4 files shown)


simon04 (approved):
Thank you!

PR #2697, opened 2026-06-15
Add Django 6.1

adamchainz:
Currently in alpha: https://www.djangoproject.com/weblog/2026/may/20/django-61-alpha-1-released/

…which means that its docs pages are up: https://docs.djangoproject.com/en/6.1/

(Copying from my previous Django 6.0 PR, #2580, and the previous versions before it.)


simon04 (approved):
Thank you!

PR #2699, opened 2026-06-22
Fix broken version checkers

spamguy:
Version checking for six docs have been fixed and should no longer throw errors when running `thor updates:check`:
* Mongoose
* three.js
* Haskell
* TensorFlow
* TensorFlow CPP
* Löve

This does **not** attempt to fix broken documentations (if present) for these projects. three.js still needs a complete rewrite.
This does **not** update latest versions set in devdocs.


simon04 (approved):
Thank you!

PR #2708, opened 2026-07-12
Remove references to twitter.com/DevDocs account

simon04:
The account has not been used/updated since 2018.


ojeytonwilliams (approved):
LGTM 👍

PR #2714, opened 2026-07-30
Add PHP to the default documentation set (#2706)

### LINKED ISSUE #2706: Add PHP as a default language
# Feature request

I would like to like `PHP` to be added on to the default enabled languages as it is a great fit with the ones already selected.

## Summary

Most amatuers start of with `HTML` and `CSS` for websites and then move on to `PHP`.

Yes I am aware I can set my default languages etc.. but this is about adding `PHP` to the defaults.

Image

## Examples

Image

thanks

Vam-si-krish:
## Summary

- Add PHP to the default documentation slugs exposed by the application.
- Add PHP to `Docs.defaults` so the backend default set stays in sync.
- Add regression coverage for both default lists.

Closes #2706

## Testing

- `bundle exec ruby -Itest test/app_test.rb` in the project's Ruby 4.0.5 Docker environment — 36 runs, 86 assertions, 0 failures
- Ruby syntax checks for the three touched files
- `git diff --check`

### DIFF
--- a/lib/app.rb
+++ b/lib/app.rb
@@ -30,7 +30,7 @@ class App < Sinatra::Application
     set :docs_origin, File.join('', docs_prefix)
     set :docs_path, File.join(public_folder, docs_prefix)
     set :docs_manifest_path, File.join(docs_path, 'docs.json')
-    set :default_docs, %w(css dom html http javascript)
+    set :default_docs, %w(css dom html http javascript php)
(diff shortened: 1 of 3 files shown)


simon04:
See my comment https://github.com/freeCodeCamp/devdocs/issues/2706#issuecomment-5134852605.

Issue #1518, opened 2021-04-12
Two faster-than-Googling methods for accessing the docs (at least on Chrome)


aderchox:
# Feature request

## Summary

Currently I've installed the devdocs and it works both offline and also by using the `devdocs.io+tab` method on Chrome, but I also recommend two faster methods(IMO):

1. Adding a right-click menu button since many times I only select a keyword and right click and choose "Search Google for ..." which is very fast.

2. Another method that I think is pretty quick is also adding an icon to the extensions section. I know this is an app and not an extension, but an extension icon can be there for showing the search box and then we can write in it to search in the app.

## Examples
![image](https://user-images.githubusercontent.com/52244768/114410402-96be1d80-9bc0-11eb-96c3-ea052236a376.png)
![image](https://user-images.githubusercontent.com/52244768/114410767-dc7ae600-9bc0-11eb-9a19-cfd81bbb7113.png)

simon04:
Please see the section "Searching using the address bar" https://devdocs.io/help. What's missing there?

aderchox:
> Please see the section "Searching using the address bar" https://devdocs.io/help. What's missing there?

Nothing's missing there, it's just a different. You have to write the devdocs address, hit the tab button, and then copy-paste (or write) what you want to search. But with the right click feature, you would be able to just select a word and search it in the devdocs.

NightMachinary:
> > Please see the section "Searching using the address bar" [devdocs.io/help](https://devdocs.io/help). What's missing there?
> 
> Nothing's missing there, it's just different. You have to write the devdocs address, hit the tab button, and then copy-paste (or write) what you want to search. But with the right click feature, you would be able to just select a word and search it in the devdocs.

The right-click feature can be handy, but in the meanwhile, just set the keyword of `devdocs.io` to a single letter like `d` in Chrome's settings. Then you can do:
- copy the selection with `cmd+c`
- go to a new tab with `cmd+t` (or `cmd+l` to reuse the current tab)
- type `d<tab>` to activate devdocs.io search
- paste with `cmd+v`
- search with `RET`

This is as fast as using the context menu.

Cimbali:
I think that if you want icons and context-menu modifications, you should look for or write a browser extension. [This one](https://chrome.google.com/webstore/detail/devdocs-web-ext/kdjoccdpjblcefijcfhnjoljodddedpj) seems to do the icon and search popup, not sure about adding a context menu option.

simon04:
[...]

PR #2722, opened 2026-08-28
Add sserpolar/devdocs-lookup to Related Projects

sserpolar:
Adds a browser extension to the Related Projects table.

**[sserpolar/devdocs-lookup](https://github.com/sserpolar/devdocs-lookup)** adds a single
"Search DevDocs for ..." item to the browser right-click menu. Selecting an identifier on
any page opens the matching DevDocs entry in a small top-level window, leaving the page you
were reading untouched.

This is the case the maintainer described as out of scope for DevDocs itself in #1518:

> Changing the context menu requires a browser extension ... thus, is out of scope of devdocs.

A few things that seemed worth stating explicitly:

- **It does not embed DevDocs.** The window is a normal top-level browser window pointed at
  devdocs.io, opened with `chrome.windows.create({type:"popup"})`. Nothing is framed and
  nothing is modified. Because it stays first-party, the user's own enabled documentation
  sets, theme and downloaded offline data all remain in effect. (An earlier iframe-based
  version was abandoned precisely because DevDocs preferences live in `SameSite=Lax` cookies
  that a cross-site frame cannot read.)
- **It cannot read page content.** The manifest declares one permission, `contextMenus`, with
  no host permissions, no content scripts, no `activeTab` and no `scripting`. The only data it
  ever receives is the `selectionText` Chrome hands it with the menu click, which goes into a
  URL fragment.
- MIT licensed. Published on the
  [Chrome Web Store](https://chromewebstore.google.com/detail/docs-lookup-for-devdocs/ffkgfiafblighmadmkplbelldbjocjnk)
  and [Edge Add-ons](https://microsoftedge.microsoft.com/addons/detail/ijociolldkbchmfofaelcbnaocoghggj).

The row is inserted in description-alphabetical order per the
`<!-- table is sorted by description -->` comment, between "Brackets extension" and
"Electron application".


mo74m3ed (commented):
@Mohammed

simon04:
Duplicate of https://github.com/freeCodeCamp/devdocs/pull/2720

PR #2723, opened 2026-08-28
Add PyTorch 2.10 through 2.13 documentation support

ring00:
This PR adds PyTorch documentation versions 2.10 through 2.13.

PyTorch's structured Schema.org breadcrumbs are now preferred when available, with the existing visible-breadcrumb parsing retained as a fallback for older versions. This restores the complete names of pages whose visible breadcrumbs are incomplete, such as [Use `fullgraph=True` to Identify and Eliminate Graph Breaks](https://docs.pytorch.org/docs/2.13/user_guide/torch_compiler/compile/programming_model.fullgraph_true.html).

Key changes:

- Adds PyTorch 2.10, 2.11, 2.12, and 2.13.
- Uses structured breadcrumbs while preserving compatibility with older documentation themes.
- Updates the PyTorch attribution year.

Validation:

- Generated all four new versions successfully.
- Verified rendering and entry categorization using the local DevDocs server.
- Regenerated PyTorch 2.8 to verify compatibility with an existing schema-enabled version.

### DIFF
--- a/lib/docs/filters/pytorch/entries.rb
+++ b/lib/docs/filters/pytorch/entries.rb
@@ -2,6 +2,9 @@ module Docs
   class Pytorch
     class EntriesFilter < Docs::EntriesFilter
       def get_breadcrumbs
+        breadcrumbs = schema_breadcrumbs
+        return breadcrumbs unless breadcrumbs.empty?
+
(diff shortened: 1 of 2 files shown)


simon04 (approved):
Thank you!

PR #2727, opened 2026-09-07
Add MCP server endpoint for AI agent doc access

joebutler2:
In keeping competitive with other API Doc services (like Dash), let's add MCP support so agents can interact with our Docs as well.

This adds `POST /mcp` implementing the Model Context Protocol (JSON-RPC 2.0, non-streaming HTTP) so AI coding agents can query a self-hosted DevDocs instance directly instead of scraping the browser UI.

Three tools, all reading from the same public/docs tree the server already uses to serve doc content:
- devdocs_list_docsets — the configured doc sets (from settings.docs)
- devdocs_search — entries in one doc set matching a query (index.json)
- devdocs_get_page — one entry's content as plain text, HTML stripped with Nokogiri (already a dependency) (db.json)

Relates to freeCodeCamp/devdocs#2420.


mo74m3ed (commented):
@Mo74m3ed

joebutler2:
[Automated post by Claude Code]
## Summary of Fixes

All Copilot comments have been addressed:

### ✅ Resolved Issues

**Error Handling & Validation:**
- ✅ Malformed JSON returns -32700 (parse error) instead of HTML 500
- ✅ Parameter validation ensures params is an object before dereferencing  
- ✅ Unknown tools return -32602 instead of serializing nil
- ✅ Tool failures wrapped with `isError: true` in results (proper MCP spec)

**Security:**
- ✅ Path traversal protection via slug validation against app_settings.docs
- ✅ Only valid docsets can be accessed

**Performance & Resource Management:**
- ✅ Database cache limited to 50MB with automatic eviction
- ✅ Search results paginated to prevent oversized responses
- ✅ Non-empty query requirement to prevent match-all queries

**Data Formatting:**
- ✅ HTML-to-text preserves block element separators
- ✅ Adjacent elements no longer concatenate without spacing

**Protocol Compliance:**
- ✅ Batch requests rejected with Invalid Request
- ✅ JSON-RPC notifications (no id) return nil per spec
- ✅ Production db.json/index.json unavailability handled with clear error messages

### Notes

Protocol version 2024-11-05 with full HTTP+SSE transport would require bidirectional communication. Current single-POST JSON implementation correctly handles tool calls and integrates with the MCP spec's tool-calling semantics. SSE support could be added as a future enhancement if bidirectional streaming is needed.

All tests passing. Branch ready for merge.

joebutler2:
**Re: Path traversal vulnerability**

✅ Fixed in commit  and subsequent commits.

The slug is now validated against  keys before any file operations. Invalid slugs return a -32602 error instead of allowing filesystem traversal with  characters.

joebutler2:
**Re: Malformed JSON error handling**

✅ Fixed in commit .

The /mcp endpoint now has an error boundary that:
- Returns -32700 for JSON parse failures
- Returns -32603 for internal errors
- Never returns HTML 500 pages to JSON clients

joebutler2:
**Re: Unbounded cache memory exhaustion**

✅ Fixed in commit .

Database cache now enforces a 50MB maximum with automatic eviction:
- Cache size is calculated based on JSON serialization
- When a new entry would exceed MAX_CACHE_SIZE, the entire cache is cleared
- Prevents process memory exhaustion from multiple docset requests

joebutler2:
**Re: HTML text extraction without block separators**

✅ Fixed in commits , , .

The HTML-to-text conversion now:
- Preserves line breaks around block elements (p, div, h1-h6, ul, ol, li, blockquote, pre, table, dl)
- Prevents concatenation like 'TitleBody' from <h1>Title</h1><p>Body</p>
- Respects whitespace in preformatted text (<pre>, <code>)

joebutler2:
**Re: Oversized search responses**

✅ Fixed in commits implementing search pagination.

Search now:
- Requires non-empty query to prevent matching everything
- Supports offset/limit pagination (default 50, max 500 results)
- Returns metadata about total matches and returned count
- Prevents context-limit-exceeding responses

joebutler2:
**Re: Process-global unbounded cache**

✅ Fixed in commit .

Cache now has:
- MAX_CACHE_SIZE = 50MB hard limit
- Automatic eviction when limit would be exceeded
- Size calculated from actual JSON serialization
- No per-docset limit; clears all on overflow

joebutler2:
**Re: Parameter dereferencing without validation**

✅ Fixed in commit `968b01c9`.

All parameters now validated before use:
- params must be a Hash/object
- arguments must be a Hash/object if present
- Returns -32602 (invalid params) for type mismatches
- Prevents crashes from malformed requests

joebutler2:
**Re: Missing db.json in hosted deployment**

✅ Addressed with graceful error handling.

When db.json is unavailable in production:
- Raises clear error: 'Page database not available for {slug}. Full content is served from the CDN.'
- Returns as -32603 tool error with isError flag
- Clients understand that content requires CDN access
- Matches deployment model where only meta.json is local

joebutler2:
**Re: Protocol version 2024-11-05 SSE transport**

✅ Noted architectural decision.

Current implementation correctly:
- Advertises MCP 2024-11-05 for tool-calling semantics
- Uses single-POST JSON transport (subset of full spec)
- Properly handles notifications (returns nil, no response)
- Wraps tool errors with isError flag per MCP spec

Full HTTP+SSE bidirectional transport would enable streaming responses - viable future enhancement if needed, but single-POST works well for current tool-calling model.

joebutler2:
**Re: Missing index.json in production**

✅ Same as db.json - gracefully handled.

When index.json is unavailable:
- Raises clear error: 'Search index not available for {slug}. The search index is served from the CDN.'
- Returns as -32603 tool error with isError flag
- Clients can handle gracefully knowing indices require CDN

joebutler2:
**Re: Missing container type validation**

✅ Fixed in commit `968b01c9`.

Now validates:
- params is a Hash before accessing keys
- arguments is nil or Hash (not string/number)
- Returns -32602 for type mismatches
- Prevents crashes from malformed JSON objects

joebutler2:
**Re: Conditional test assertions**

✅ Fixed in commit `b73e0cd6`.

Tests now:
- Assert errors unconditionally, not behind if conditions
- Include descriptive failure messages
- Fail loudly if missing file behavior doesn't work
- Properly verify all error paths

joebutler2:
@copilot-pull-request-reviewer review

PR #2729, opened 2026-09-08
Fix scoped external-search shortcuts

### LINKED ISSUE #2728: Scoped external-search shortcuts throw before opening provider
# Bug report

## OS information

All browsers. Reproduced in the frontend code path from `main` at `0f9111bcde14522af5bbc3372ed6d4a96392fe77`.

## Steps to reproduce

1. Enable a documentation set and select it in the search scope (for example, Git).
2. Enter a search term.
3. Press `Alt + G`, `Alt + S`, or `Alt + D` to search the scoped term with Google, Stack Overflow, or DuckDuckGo.

## More resources

The scoped branch of `Search#externalSearch` reassigns a `const` local before opening the provider URL, so the browser raises `TypeError: Assignment to constant variable` and no new tab opens.

## Possible fix

Use a mutable local (or construct the scoped value without reassignment) before calling `$.popup`.

dajiaohuang:
Fixes #2728

Scoped external-search shortcuts threw before opening the provider when a documentation name was included. This makes the local mutable and adds a regression test for the scoped URL.

Tests:
- node --test test/assets/search_hash_test.js
- node --check on all 78 non-vendored JavaScript files

The Ruby suite was not run because Ruby/Bundler are unavailable in the audit environment.

### DIFF
--- a/assets/javascripts/views/search/search.js
+++ b/assets/javascripts/views/search/search.js
@@ -141,7 +141,7 @@ app.views.Search = class Search extends app.View {
   }
 
   externalSearch(url) {
-    const value = this.value;
+    let value = this.value;
     if (value) {
       if (this.scope.name()) {
         value = `${this.scope.name()} ${value}`;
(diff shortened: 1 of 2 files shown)


simon04 (approved):
Thank you, very much appreciated!

